PR #3, opened 2018-01-24
The Epstein Civil violence model was modified to include sliders and …

Zazo100:
We modified the Epstein Civil Violence example to include sliders and result within the visual environment. 
Secondly we fixed a small mistake where the cop vision variable was always set to 1 in the code.
Lastly we slightly changed the portrayal of the 'cops' so they are easier to distinguish

This was done as part of the course 'Agent-Based Modeling' at the 'University of Amsterdam' (UvA).

Thank you for this basis code (and Mesa in general), for our project we shall implement some automatic variation in cops density and Legitimacy and see how this affects the outputs. But this is no longer part of the Epstein example.


jackiekazil:
We didn't realize people were using this repo. There is a lot out of date. Please submit to this repo: https://github.com/projectmesa/mesa/tree/master/examples

PR #10, opened 2022-11-16
Move examples from mesa repo to mesa-examples repo.

jackiekazil:


CONTRIBUTING.rst
```diff
@@ -0,0 +1,86 @@
+Contributing
```
jackiekazil, line 1: This file was thrown together and probably needs future iterations.

CONTRIBUTING.rst
```diff
@@ -0,0 +1,86 @@
...
+.. _`an issue` : https://github.com/projectmesa/mesa-examples/issues
+
+**To submit a contribution**
+
+- Create a ticket for the item that you are working on.
+- Fork the mesa-examples repository.
+- `Clone your repository`_ from Github to your machine.
+- Create a new branch in your fork: ``git checkout -b BRANCH_NAME``
+- Run ``git config pull.rebase true``. This prevents messy merge commits when updating your branch on top of Mesa main branch.
+- If you are using the existing Mesa version that a model might be pinned to, then you will want to run ``pip install -r requirements.txt`` from inside the example folder you are working with. If you are making changes using the main repo or other version, do not install the version in the requirements file & make sure to update if the example is pinned.
```
jackiekazil, line 28: This line specifically could use some finessing. People with less Python experience are more likely to contribute a model than core code. But here is a situation where they need to pay attention to the versions of things. That is what I was trying to describe. Maybe we pass this thru for now, but reapproach when we do a rewrite of this as I suggest here: https://github.com/projectmesa/mesa/issues/963#issuecomment-1336722683

examples/Virus_on_Network/requirements.txt
```diff
@@ -0,0 +1,2 @@
+networkx>=2.0
+mesa
```
jackiekazil, line 2: @rht this line is specifically added in the move -- because this model was failing.

jackiekazil:
@rht when you go through this - can you make sure that recent PR changes made it in. I believe I captured them, but would like a second set of eyes on this explicitly, because I did have a hiccup in the process.

examples/Boltzmann_Wealth_Model/Readme.md
```diff
@@ -2,33 +2,33 @@
 
 ## Summary
 
-A simple model of agents exchanging wealth. All agents start with the same amount of money. Every step, each agent with one unit of money or more gives one unit of wealth to another random agent. This is the model described in the [Intro Tutorial](http://mesa.readthedocs.io/en/latest/intro-tutorial.html).
+A simple model of agents exchanging wealth. All agents start with the same amount of money. Every step, each agent with one unit of money or more gives one unit of wealth to another random agent. This is the model described in the [Intro Tutorial](https://mesa.readthedocs.io/en/latest/tutorials/intro_tutorial.html).
```
rht, line 5: Confirmed that this is the change done in https://github.com/projectmesa/mesa/commit/238c2c0bd6286970418bd1439bf0c136df665782.

examples/Virus_on_Network/requirements.txt
rht: I was about to say that this should have been pinned to `mesa==1.x`, because people might be cloning this repo and forget to update the repo when it has been changed to be compatible with 2.0 instead of 1.x. But the diff in this PR is huge, and any small iterations on top of it will slow down the process. And so, I'm merging now.

PR #12, opened 2022-12-05
requirements: Pin Mesa version to 1.x

rht:


Corvince:
You are correct with everything related to pip version management, but this PR is fine. Pinning to 1.y is correct, since increasing the minor version number (the y) must not include breaking changes. 

So for example new features might be added, but nothing backwards incompatible. That would require a version 2.y

valer-cara:
Indeed, however 1.2.(1?) minor version update seems to introduce some [breaking changes](https://github.com/projectmesa/mesa/commit/a384a0dfd44ae869017d127ef3049c2cffc91cbb) by making `model.space.Grid` private (renaming to _Grid) which threw some errors, for example Conway with mesa 1.2.1, which is fixed by moving to 1.1.

```
AttributeError: module 'mesa.space' has no attribute 'Grid'. Did you mean: '_Grid'?
```

Corvince:
Good catch! We should probably release a bug fix release where Grid is only deprecated, not removed. @projectmesa/maintainers

jackiekazil:
LGTM to me.

PR #14, opened 2022-12-23
add the example 'caching_and_replay' which uses Mesa-Replay

Logende:


examples/caching_and_replay/README.md
```diff
@@ -0,0 +1,42 @@
+# Caching and Replaying Schelling
+
+## Summary
+
+This example applies caching on the Mesa [Schelling example](https://github.com/projectmesa/mesa-examples/tree/main/examples/Schelling).
+It enables a simulation run to be "cached" or in other words recorded. The recorded simulation run is persisted on the local file system and can be replayed at any later point.
+
+It uses the [Mesa-Replay](https://github.com/Logende/mesa-replay) library and puts the Schelling model inside a so-called `CachableModel` wrapper that we name `CachableSchelling`.
+From the outside perspective, the new model behaves the same way as the original Schelling model, but additionally supports caching.
```
rht, line 9: Should be "from the user's perspective".

rht:
I tried to lint the README.md via https://vale.sh/, as an experiment. I found that it said "cachable" is not a thing. According to https://github.com/GeneaLabs/laravel-model-caching/issues/159, "cacheable" occurs ~9x more often than "cachable". And so, I think we should use "cacheable".

As a side note, I think we should have a prose linter for the repo. Haven't decided yet which one to use. Contenders: https://github.com/errata-ai/vale#functionality.

Corvince:
So cool to see this. It is always nice to see libraries coming up to populate the mesa ecosystem! I am excited to take a closer look at the library and am positive about this PR being mergeable.

examples/caching_and_replay/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Installation
+
+To install the dependencies use pip and the requirements.txt in this directory. e.g.
+
+```
+    $ pip install -r requirements.txt
```
rht, line 16: @Logende you are already aware of this, but just a reminder to add Mesa-Replay in the requirements.txt.

rht:
Pressed the wrong button. Wanted to say that one more use case would be to resume execution of a simulation run at a checkpoint.

examples/caching_and_replay/README.md
Logende: changed

Logende:
> I tried to lint the README.md via https://vale.sh/, as an experiment. I found that it said "cachable" is not a thing. According to [GeneaLabs/laravel-model-caching#159](https://github.com/GeneaLabs/laravel-model-caching/issues/159), "cacheable" occurs ~9x more often than "cachable". And so, I think we should use "cacheable".
> 
> As a side note, I think we should have a prose linter for the repo. Haven't decided yet which one to use. Contenders: https://github.com/errata-ai/vale#functionality.

Renamed cachable to cacheable in both the mesa-replay repo and in this example.

examples/caching_and_replay/README.md
Logende: done

Logende:
> Pressed the wrong button. Wanted to say that one more use case would be to resume execution of a simulation run at a checkpoint.

I will add something like a "go_to_step(step: int)" function and a step_backwards() function to cacheable_model. This could then be integrated into the UI to allow the user to jump to some step or move backwards.

rht:
> > Pressed the wrong button. Wanted to say that one more use case would be to resume execution of a simulation run at a checkpoint.
> 
> I will add something like a "go_to_step(step: int)" function and a step_backwards() function to cacheable_model. This could then be integrated into the UI to allow the user to jump to some step or move backwards.

No worries about this one for this PR. I will open an issue in https://github.com/Logende/mesa-replay.

Logende:
> > > Pressed the wrong button. Wanted to say that one more use case would be to resume execution of a simulation run at a checkpoint.
> > 
> > 
> > I will add something like a "go_to_step(step: int)" function and a step_backwards() function to cacheable_model. This could then be integrated into the UI to allow the user to jump to some step or move backwards.
> 
> No worries about this one for this PR. I will open an issue in https://github.com/Logende/mesa-replay.

Already implemented and tested :soil
https://github.com/Logende/mesa-replay/commit/ff94c068ff3042f0aebbf0b4580f2ed93ea28972

rht:
cc: @snunezcr. Finally, we are having a model replay feature!

examples/caching_and_replay/cacheablemodel.py
```diff
@@ -0,0 +1,28 @@
+from model import Schelling
+from mesa_replay import CacheableModel, CacheState
+
+
+class CacheableSchelling(CacheableModel):
+    """A wrapper around the original Schelling model to make the simulation cacheable and replay-able.
+    Uses CacheableModel from the Mesa-Replay library, which is a decorator that can be put around any regular mesa model
```
rht, line 7: I don't see any decorator used in this example.

Logende: Sorry, I don’t really understand. What do you mean you don’t see any decorator in this example? What would you expect to see instead?
From an sw architectural point of view, CacheableModel uses the decorator pattern and it is being inherited from here. Do we have a different understanding of a decorator? Is there some Python specific usage of the when decorator?

rht: >  Is there some Python specific definition of the term decorator?

It's a [Python decorator](https://en.wikipedia.org/wiki/Python_syntax_and_semantics#Decorators), and is the first definition that comes to people's mind, at least for Python users.

Logende: Ah, that’s good to know. Maybe I can use those Python decorators to improve Mesa-Replay.

rht: I'd say "wrapper" is sufficient. Saying "decorator" to mean what you meant may be fine in other languages, but it's misleading in Python.

Logende: Yeah, I agree. Changing decorator to wrapper in the description should be sufficient and fine. I was just thinking if it makes sense to actually use the Python decorator functionality to make ModelCacheable a little more accessible. This I will test out today in the evening. 

Just to name one example: Right now, ModelCacheable explicitly passes the attribute calls it gets to the model, to make it behave like the model from the user’s perspective. When using Python decorators this may no longer be needed, because then the object would still be the model itself, now just with added functionality. So Python decorators might make it possible to simplify the code a little.

rht: Python decorators are meant to modify a function/class on a very specific/limited scope. If `ModelCacheable` were to be made a decorator, it would add too much functionality. So far we have `run_model_until_condition_met`, but who knows what features are added later. This is debatable, though.

I think `CacheableModel` is similar to Mesa's `batch_run`, in that it operates on a model object.

examples/caching_and_replay/README.md
```diff
@@ -0,0 +1,42 @@
+# Caching and Replaying Schelling
```
rht, line 1: Should be "Schelling Model with Caching and Replay".

examples/caching_and_replay/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+To run the model interactively, run ``mesa runserver`` in this directory. e.g.
+
+```
+    $ mesa runserver
+```
+
+Then open your browser to [http://127.0.0.1:8521/](http://127.0.0.1:8521/) and press Reset, then Run.
+
+After running a regular simulation, you can **replay** your latest simulation run by first enabling the Replay switch and then pressing Reset.
```
rht, line 29: Do I read this as:
- Run regular simulation with "Replay last run?" as False
- Then set "Replay last run?" to True
- Press "Reset"
- Start the simulation

What would happen if I set the replay switch to `True` in the beginning, while the cache is still nonexistent? Either throw an error, or implicitly run and save to cache. Either way, the user needs to be informed of the cache existence/non-existence.

Also, I think it should be called "Replay cached run?" instead. Because "last run" implies most recent run, which is not always the case.

Logende: You understand the steps correctly. 

If the replay switch is set to `True` in the beginning, the user would get a "cache file not existing" error message. This would happen in the console though and not in the web view. Is there a simple way to inform the user about the cache file existence in the web view?

It would be okay for me to rename `Replay cached run` into `Replay cached run`. I think it should always be the case that this is the latest run though: when we simulate and do not replay, a new run is executed but will in any case be written to the cache. If there is a replay, it will replay the run from cache, which is the same as the latest simulation run or the latest replay.

examples/caching_and_replay/README.md
Logende: changed

examples/caching_and_replay/README.md
rht: You can display custom information as a text, e.g. https://github.com/projectmesa/mesa-examples/blob/37e0a9de1e8c7ee9c907757d744053606539744c/examples/Schelling/server.py#L10. In this case, it would be to show whether the cache file exists.

> After running a regular simulation

You have to make it explicit that the user has to disable the switch for the first run.

>  If there is a replay, it will replay the run from cache, which is the same as the latest simulation run or the latest replay.

So, caching always happens whenever the server is not in replay mode. This needs to be explicitly told via a short explanation, possibly near the text info showing whether the cache exists or not. Anyway, "replay cached run" is still safer, because there is a possibility that the user will save the cache somewhere else, and to replay it on a different machine, etc.

Logende: I've added explicit information about the caching status and that replay can be used only if a cache file already exists now.

examples/caching_and_replay/README.md
```diff
@@ -26,7 +26,7 @@ To run the model interactively, run ``mesa runserver`` in this directory. e.g.
 
 Then open your browser to [http://127.0.0.1:8521/](http://127.0.0.1:8521/) and press Reset, then Run.
 
-After running a regular simulation, you can **replay** your latest simulation run by first enabling the Replay switch and then pressing Reset.
+After running a regular simulation, you can **replay** your latest cached simulation run by first enabling the Replay switch and then pressing Reset.
```
rht, line 34: Still no explicit statement that replay needs to be disabled for the first run. You should rephrase to describe the first run on an imperative statement.

Something like:
"First, run the simulation with the Replay switch disabled, ...."

examples/caching_and_replay/run.py
```diff
@@ -9,12 +11,23 @@
...
 # As 'replay' is a simulation model parameter in this example, we need to make it available as such
-model_params["replay"] = mesa.visualization.Checkbox("Replay last run?", False)
+model_params["replay"] = mesa.visualization.Checkbox("Replay cached run?", False)
+
+
+def get_cache_file_status(_):
+    """
+    Display an informational text about caching and the status of the cache file (existing versus not existing)
+    """
+    return f"This example writes every simulation run into a cache file. Cached runs can be replayed later. " \
```
rht: The description is too long for me, but I think is also fine if you don't want to make it shorter.

examples/caching_and_replay/README.md
rht: This is the only remaining necessary change I requested. Everything else looks good to me.

Logende: Okay, fixed now.

examples/caching_and_replay/run.py
Logende: I made it a little shorter now

Tortar:
I want just to say that we should try to merge `mesa-replay` inside `mesa` itself in the future, because it can be a very useful functionality so for users to benefit the most it should be in the main repo :-)

rht:
> I want just to say that we should really try to merge `mesa-replay` inside `mesa` itself in the future overcoming the last difficulties, because it can be a very useful functionality, for users to benefit the most it should be in the main repo :-)

That's the plan. Currently, it's a separate repo so that it's more convenient to do breaking changes. Also, the cache size is too big at the moment.

@jackiekazil any comments?

rht:
Merging now. Thank you @Logende !

PR #17, opened 2023-01-26
add gis examples

wang-boyu:
For https://github.com/projectmesa/mesa-geo/issues/128.

Moved the following examples into `gis/` folder:
- https://github.com/projectmesa/mesa-geo/tree/main/examples
- https://github.com/wang-boyu/agents-and-networks-in-python


wang-boyu:
Good question. Let me figure this out.

wang-boyu:
- Update dependencies of Mesa-Geo to be ~=0.5.0
- Zip data files in the `agents_and_networks` example
- Update data licenses in the following examples to be CC BY-SA 4.0 as mentioned in the [abmgis repo](https://github.com/abmgis/abmgis/blob/master/LICENSE.md) from where we got the data:
  * agents_and_networks
  * population
  * rainfall
  * urban_growth

Should be ready for review now.

tpike3:
As discussed this is a big merge, but the license issue is resolved. So it would be prudent to merge and then deal with any issues as we get mesa-examples stood up.

PR #18, opened 2023-01-29
Add Sugarscape with traders example

tpike3:
- Add sugarcape_g1mt that is consistent with complexity tutorial
- Include interactive version with single run, batch run and server options
- Update .gitignore to ignore IDE environment files


.gitignore
```diff
@@ -55,3 +55,31 @@ docs/_build/
...
+# Apple OSX
+*.DS_Store
+
+# mypy
+.mypy_cache/
+.dmypy.json
+dmypy.json
+
+# pythonenv
+pythonenv*/
```
rht, line 81: This seems to be not a generic pattern to be ignored.

.gitignore
```diff
@@ -55,3 +55,31 @@ docs/_build/
...
+
+# JS dependencies
```
rht, line 83: These are specific to the Mesa repo only.

examples/sugarscape_g1mt/run.py
```diff
@@ -0,0 +1,101 @@
+import sys
+import pandas as pd
+import matplotlib.pyplot as plt
+import networkx as nx
+import mesa
+from sugarscape_g1mt.model import SugarscapeG1mt
+from sugarscape_g1mt.server import server
+
+
+#Analysis
```
rht, line 10: This needs to be `# Analysis`. It means the file hasn't been Black-ed yet.

examples/sugarscape_g1mt/Readme.md
```diff
@@ -0,0 +1,90 @@
...
+
+```
+    $ mesa runserver
+```
+
+Then open your browser to [http://127.0.0.1:8521/](http://127.0.0.1:8521/) and press Reset, then Run.
+
+## Files
+
+* ``sugarscape_g1mt/trader_agents.py``: Defines the SsAgent, and Sugar agent classes.
```
rht, line 81: Should be `Trader`.

examples/sugarscape_g1mt/sugarscape_g1mt/trader_agents.py
```diff
@@ -0,0 +1,352 @@
...
+
+        # convert index to positions of those cells
+        candidates = [neighbors[i] for i in candidate_indices]
+
+        min_dist = min(get_distance(self.pos, pos) for pos in candidates)
+
+        final_candidates = [pos for pos in candidates
+                            if math.isclose(get_distance(self.pos, pos), min_dist, rel_tol=1e-02
+                                            )]
+        self.random.shuffle(final_candidates)
```
rht, line 298: Reposting https://github.com/projectmesa/mesa/pull/1290#discussion_r1060319796. Not critical, but adding the comment here just in case.
```python
final_candidate = self.random.choice(final_candidates)
self.model.grid.move_agent(self, final_candidate)
```
is clearer.

examples/sugarscape_g1mt/sugarscape_g1mt/model.py
```diff
@@ -0,0 +1,168 @@
...
+            agent_id += 1
+
+    def randomize_traders(self):
+        '''
+        helper function for self.step()
+
+        puts traders in randomized list for step function
+        '''
+
+        trader_shuffle = list(self.schedule.agents_by_type[Trader].values())
```
rht, line 125: It would make more sense with `traders_shuffle`.

.gitignore
tpike3: My understanding is this is the files for the python environments, are we sure don't want to ignore this

.gitignore
tpike3: Right, will delete

examples/sugarscape_g1mt/run.py
tpike3: Hmmm...not sure why black didn't run, but it is fixed now

examples/sugarscape_g1mt/Readme.md
tpike3: Will fix

.gitignore
rht: Python virtualenv folder names are arbitrary. You can't expect anyone to name their folder "pythonenv".

tpike3: I will delete it but we should also delete it form the main repo

rht: Yes indeed. It was added in https://github.com/projectmesa/mesa/commit/0f1bcdc0f441b13a0b4dfd9858d7c63bc67f56bc.

examples/sugarscape_g1mt/sugarscape_g1mt/model.py
```diff
@@ -0,0 +1,168 @@
...
+  '''
+  return [item for sublist in list_of_lists for item in sublist]
+
+def geometric_mean(list_of_prices):
+  '''
+  find the geometric mean of a list of prices
+  '''
+  return np.exp(np.log(list_of_prices).mean())
+
+def get_trade(agent):
```
rht, line 23: Just a note, not a change request: it'd be clearer with `get_trade_partners`, but if this is renamed, it would be harder for people who come from the CE video to associate `get_trade` with `get_trade_partners`.

.gitignore
tpike3: fixed

examples/sugarscape_g1mt/sugarscape_g1mt/model.py
tpike3: Fair point

rht:
The trader.png's license is CC BY-NC-SA, which requires attribution. Also, I think its size is way too big (315 KB).

tpike3:
Removed trader and just made it a circle

rht:
LGTM

PR #19, opened 2023-03-25
Fix eat function in Sugrascpae_g1mt

tpike3:
This fixes the complexity explorer user identified error as well as two other bugs found in running the example (processing batch run and running batch run)


examples/sugarscape_g1mt/sugarscape_g1mt/trader_agents.py
```diff
@@ -322,23 +322,24 @@ def move(self):
...
     def eat(self):
-        """
-        Function for trader to consume sugar and spice in grid cell
-        """
-
         # get sugar
         sugar_patch = self.get_sugar(self.pos)
-        # eat sugar
+
         if sugar_patch:
```
rht, line 328: Shouldn't this be:
```python
if sugar_patch:
    self.sugar  += sugar_patch.amount
    sugar_patch.amount = 0
self.sugar -= metabolism_sugar?
```

tpike3: I interpreted the matrix conversation that you didn't have a strong opinion either-- the way I changed it or the way you commented. 

With how I changed it is just adding lines, although more verbose, vs rewriting the if statement

What do you think?

rht: I do think simplifying to not have else does definitely clarify the code that `self.sugar -= metabolism_sugar` happens in all branches of execution. The one that I don't have opinion is on the grouping collect sugar, eat sugar, collect spice, eat spice vs collect sugar & spice, eat sugar & spice.

tpike3: Ahhh...got it I will update and sync the tutorial as well.

PR #22, opened 2023-04-24
Fix Capitalization of File Names

ItsQuinnMoore:
Change example file names to all lowercase and make the README uppercase.


jackiekazil:
LGTM

jackiekazil:
Sorry all, I accidentally merged before a squash, and I thought a revert might fix it (#23), but I think that is going to add more noise so I am going to leave it as is.

PR #25, opened 2023-04-24
Update outdated readme for Boltmann example + trim requirements.txt

houssam7737:
Update outdated readme to be more accurate + clean up requirements.txt of unnecessary dependencies.

Script runs succesfully and I could see the simulation on my browser (pretty cool!).

One question: there is current a discrepancy between the jupyter notebook and the code in model.py. Notably the `self.datacollector.collect(self)` happens both within the init and in the step function in `model.py`, and the order is also switched within `step` (does it matter when the data collector collects data? would it miss the last step in the jupyter notebook?)

Excited for my first PR/contribution!

### DIFF
--- a/examples/boltzmann_wealth_model/Readme.md
+++ b/examples/boltzmann_wealth_model/Readme.md
@@ -2,18 +2,26 @@
 
 ## Summary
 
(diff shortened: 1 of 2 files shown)


jackiekazil:
LGTM YAY!

PR #26, opened 2023-04-24
Use Pathlib

catherinedevlin:
Pathlib is a new(-ish) standard library module (introduced 3.4) that provides a cleaner and more reliable way to handle file paths than the old os.path methods. They bake in OS-safety, adapting to Windows back-slashes and Unix forward-slashes seamlessly. Switching to Pathlib cleans up the code a little and avoids possible Windows problems.

Code by Phil Robare (@versilimidude2)

PR #27, opened 2023-04-24
Use pre-commit

catherinedevlin:
- Added pre-commit configuration
- Applied black formatting

Since the CI is set up to run format checks - see the failures at https://github.com/projectmesa/mesa-examples/pull/26 - I thought the repo overall could use similar pre-commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,3 +79,5 @@ dmypy.json
 
 # Virtual environment
 venv/
+
+examples/caching_and_replay/my_cache_file_path.cache
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,6 @@ repos:
 -   repo: https://github.com/psf/black
     rev: 23.3.0
     hooks:
-      - id: black
       - id: black-jupyter
 -   repo: https://github.com/asottile/pyupgrade
     rev: v3.3.1
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,21 @@
+ci:
+    autoupdate_schedule: 'monthly'
+    autofix_prs: false
+
+repos:
+-   repo: https://github.com/psf/black
+    rev: 23.3.0
+    hooks:
+      - id: black
+      - id: black-jupyter
+-   repo: https://github.com/asottile/pyupgrade
+    rev: v3.3.1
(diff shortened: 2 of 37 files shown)


gis/geo_sir/model.py
```diff
@@ -101,7 +104,8 @@ def step(self):
         self.steps += 1
         self.reset_counts()
         self.schedule.step()
-        self.space._recreate_rtree()  # Recalculate spatial tree, because agents are moving
+        self.space._recreate_rtree()  # Recalculate spatial tree,
```
rht, line 107: The comment should be before the `self.space._recreate_rtree()` line, so that there is continuity in the comment.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,21 @@
+ci:
+    autoupdate_schedule: 'monthly'
+    autofix_prs: false
+
+repos:
+-   repo: https://github.com/psf/black
+    rev: 23.3.0
+    hooks:
+      - id: black
+      - id: black-jupyter
```
rht, line 10: I think `black-jupyter` supersedes `black`. See also: https://github.com/projectmesa/mesa/blob/34eb008aa849c70f80e0fa69198d82ba2bf82193/.pre-commit-config.yaml#L9

.gitignore
```diff
@@ -79,3 +79,5 @@ dmypy.json
 
 # Virtual environment
 venv/
+
+examples/caching_and_replay/my_cache_file_path.cache
```
rht, line 83: This is the default file name in the `caching_and_replay` example, so I suppose it is fine to be added in .gitignore.

.pre-commit-config.yaml
catherinedevlin: Thanks for the suggestions; incorporated!

jackiekazil: This comment it out of date -- something is wrong with Github and it is not registering correctly. 
@catherinedevlin has fixed the issue -- it is identifiable in the file. This is super weird behavior on github ... took us awhile to troubleshoot.

gis/geo_sir/model.py
jackiekazil: changes fix this.

jackiekazil:
This looks to me.

PR #28, opened 2023-04-25
Fix bug boltzman model agent giving money to itself

houssam7737:
Pop the current agent from the list of agents on the cell, and then randomly picking an agent.


houssam7737:
pre-commit.ci autofix

jackiekazil:
LGTM

PR #33, opened 2023-05-01
Fix bug in tutorial to calculate MRS with potential trade

tpike3:
- update maybe_sell_spice() so calculate_MRS() takes potential sugar and spice based on trade
- update calculate_MRS() with kwargs so it can take potential trade but defaults to agents sugar and spice 
Update increases agent_reporter due to number of trades creating memory issues
- Update data collection so non-relevant data rmeoved on each step (i.e. sugar and spice agents don't have trade partners)

### DIFF
--- a/examples/sugarscape_g1mt/sugarscape_g1mt/model.py
+++ b/examples/sugarscape_g1mt/sugarscape_g1mt/model.py
@@ -187,6 +187,24 @@ class SugarscapeG1mt(mesa.Model):
 
         # collect model level data
         self.datacollector.collect(self)
+        """
+        Mesa is working on updating datacollector agent reporter
+        so it can collect information on specific agents from
+        mesa.time.RandomActivationByType.
+
+        Please see issue #1419 at
+        https://github.com/projectmesa/mesa/issues/1419
+        (contributions welcome)
+
+        Below is one way to update agent_records to get specific Trader agent data
+        """
+        # Need to remove excess data
+        # Create local variable to store trade data
+        agent_trades = self.datacollector._agent_records[self.schedule.steps]
+        # Get rid of all None to reduce data storage needs
+        agent_trades = [agent for agent in agent_trades if agent[2] is not None]
+        # Reassign the dictionary value with lean trade data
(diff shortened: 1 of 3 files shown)


rht:
I think we should set `sugar` and `spice` to be required arguments, making them explicit, and thus, requires less justification on whether they are correct.

tpike3:
Made calculate_MRS so it requires sugar and spice. I also increased the number of runs since the memory issue is mitigated.

examples/sugarscape_g1mt/sugarscape_g1mt/model.py
```diff
@@ -187,6 +187,13 @@ def step(self):
 
         # collect model level data
         self.datacollector.collect(self)
+        # Need to remove excess data
+        # Create local variable to store trade data
```
rht, line 191: What about https://github.com/projectmesa/mesa/issues/1419#issuecomment-1257158578 ? Which doesn't expose the internal wiring of the `DataCollector`.

tpike3: My concern with putting it in the model collector is the point of the lesson is to talk about the agent reporter. Showing the internal wiring of `datacollector`, I think is more consistent with the learning objectives as I show the source code in various lessons.. This leads back to your second comment that we would need to build a class for collecting `DataCollectorRABT`. 

So in I would prefer to show the internal wiring... I don't know.. what do you think?

rht: I suppose it is fine for now as long as there is a comment that says this part is still a work in progress, with a link to #1419.

tpike3: Comment added

rht:
Needs Black formatting,

tpike3:
> There are still trailing whitespaces in examples/sugarscape_g1mt/sugarscape_g1mt/model.py.

Sorry about that forget to install pre-commit in my mesa-examples.

rht:
LGTM

PR #35, opened 2023-05-17
feat: Create mesa-models as a package

rht:
Only `boltzmann_wealth_model` is enabled right now.
To use this, install the package, and do
```python
from mesa_models.boltzmann_wealth_model import model
model.compute_gini
```


tpike3 (approved):
LGTM

PR #36, opened 2023-05-27
Wealth model streamlit

ankitk50:
Streamlit browser app interface for Boltzmann wealth model example. Run the example using - `streamlit run app.py`. The animated grid is modelled using pandas and Altair.

### DIFF
--- a/examples/boltzmann_wealth_model/app.py
+++ b/examples/boltzmann_wealth_model/app.py
@@ -1,0 +1,117 @@
+from boltzmann_wealth_model.model import BoltzmannWealthModel
+
+import streamlit as st
+
+import time
+
+import pandas as pd
+
+import altair as alt
+
+
+model = st.title("Boltzman Wealth Model")
+num_agents = st.slider(
+    "Choose how many agents to include in the model",
+    min_value=1,
+    max_value=100,
+    value=50,
+)
+num_ticks = st.slider(
+    "Select number of Simulation Runs", min_value=1, max_value=100, value=50
+)
+height = st.slider("Select Grid Height", min_value=10, max_value=100, step=10, value=15)
+width = st.slider("Select Grid Width", min_value=10, max_value=100, step=10, value=20)
+model = BoltzmannWealthModel(num_agents, height, width)
+
+
+status_text = st.empty()
+run = st.button("Run Simulation")
+
+
+if run:
+    tick = time.time()
+    step = 0
+    # init grid
+    df_grid = pd.DataFrame()
+    df_gini = pd.DataFrame({"step": [0], "gini": [-1]})
+    for x in range(width):
+        for y in range(height):
+            df_grid = pd.concat(
+                [df_grid, pd.DataFrame({"x": [x], "y": [y], "agent_count": 0})],
+                ignore_index=True,
+            )
+
+    heatmap = (
+        alt.Chart(df_grid)
+        .mark_point(size=100)
+        .encode(x="x", y="y", color=alt.Color("agent_count"))
+        .interactive()
+        .properties(width=800, height=600)
+    )
+
+    line = (
+        alt.Chart(df_gini)
+        .mark_line(point=True)
+        .encode(x="step", y="gini")
+        .properties(width=800, height=600)
+    )
+
+    # init progress bar
+    my_bar = st.progress(0, text="Simulation Progress")  # progress
+    placeholder = st.empty()
+    st.subheader("Agent Grid")
+    chart = st.altair_chart(heatmap)
+    st.subheader("Gini Values")
+    line_chart = st.altair_chart(line)
+
+    color_scale = alt.Scale(
+        domain=[0, 1, 2, 3, 4], range=["red", "cyan", "white", "white", "blue"]
+    )
+    for i in range(num_ticks):
+        model.step()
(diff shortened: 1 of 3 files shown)


rht:
Needs to rebase the branch. You need to run `git config pull.rebase true`, then the subsequent `git pull`'s should automatically do a rebase.

ankitk50:
I get the following fail:

pyupgrade................................................................Failed
- hook id: pyupgrade
- exit code: 1
- files were modified by this hook

Not sure which change is responsible for this.

rht:
The lots of extra commits are still there. I think it's because when you did `git pull`, you pulled from your own fork. You need to ensure that the upstream repo (https://github.com/projectmesa/mesa-examples) is in your `git` config, and and do `git pull upstream main`.

examples/boltzmann_wealth_model/app.py
```diff
@@ -0,0 +1,127 @@
...
+
+import streamlit as st
+
+import time
+
+import pandas as pd
+
+import altair as alt
+
+import plotly.express as px
```
rht, line 13: Since `plotly` is no longer used, it should be removed from the import. It's a huge dependency.

examples/boltzmann_wealth_model/app.py
```diff
@@ -0,0 +1,127 @@
...
+
+from boltzmann_wealth_model.model import BoltzmannWealthModel
+
+import streamlit as st
+
+import time
+
+import pandas as pd
+
+import altair as alt
```
rht, line 11: What is the reason for switching to `altair`?

ankitk50: I could not do the animation of the grid in plotly.

examples/boltzmann_wealth_model/app.py
ankitk50: Sure

rht:
You haven't done the `git pull` after the rebase config has been enabled, as evident with the number of commits in this PR.

rht:
Can you apply [`ruff`](https://github.com/charliermarsh/ruff) and [`black`](https://github.com/psf/black) to your changes?

rht:
It's not required for this PR, but you should abstract out the code into a class, as can be seen from @wang-boyu's [implementation](https://github.com/wang-boyu/mesa-examples/blob/feat%2Fstreamlit-frontend/examples/Schelling/st_app.py).

examples/boltzmann_wealth_model/app.py
```diff
@@ -0,0 +1,117 @@
...
+
+model = st.title("Boltzman Wealth Model")
+num_agents = st.slider(
+    "Choose how many agents to include in the model",
+    min_value=1,
+    max_value=100,
+    value=50,
+)
+num_ticks = st.slider(
+    "Select number of Simulation Runs", min_value=1, max_value=100, value=50
```
rht, line 20: Should be "Select number of simulation steps".

examples/boltzmann_wealth_model/app.py
```diff
@@ -0,0 +1,117 @@
...
+
+        line = (
+            alt.Chart(df_gini)
+            .mark_line(point=True)
+            .encode(x="step", y="gini")
+            .properties(width=800, height=600)
+        )
+        line_chart.altair_chart(line)
+
+        time.sleep(0.01)
```
rht, line 108: I have noticed that the plots update are slow even without this line. Is the slowness inherent to Streamlit? Since the model is created once, it has to be due to the redraw.

examples/boltzmann_wealth_model/app.py
```diff
@@ -0,0 +1,117 @@
+from boltzmann_wealth_model.model import BoltzmannWealthModel
```
rht, line 1: The import order based on [isort](https://pypi.org/project/isort/)
```python
import time

import altair as alt
import pandas as pd
import streamlit as st

from boltzmann_wealth_model.model import BoltzmannWealthModel
```

examples/boltzmann_wealth_model/app.py
rht: Can you test if it is faster with `st.pyplot`?

rht:
The Readme.md needs to say that app.py is an optional file to run the simulation via the Streamlit interface, as an alternative to run.py. It'd be informative if app.py has the same textual description as well.

Needs to explain how to run the app.py. The Streamlit dependency shouldn't be included in requirements.txt, since this is optional. But at least the Readme.md should say what additional libraries need to be installed.

examples/boltzmann_wealth_model/app.py
ankitk50: the change is on the latest commit

ankitk50:
> The Readme.md needs to say that app.py is an optional file to run the simulation via the Streamlit interface, as an alternative to run.py. It'd be informative if app.py has the same textual description as well.
> 
> Needs to explain how to run the app.py. The Streamlit dependency shouldn't be included in requirements.txt, since this is optional. But at least the Readme.md should say what additional libraries need to be installed.

Readme has been updated accordingly.

rht:
I'm merging for now. We can iterate on the `isort` stuff, and making it faster in another PR.

PR #40, opened 2023-06-18
Bump to Mesa 2.0

rht:
And update the network examples, with a backward incompatible change in https://github.com/projectmesa/mesa/pull/1542.


tpike3 (approved):
LGTM will merge once we release 2.0

PR #47, opened 2023-07-03
solara frontend: Make play/pause work in Colab

rht:


rht:
2023-07-03-105052_304x162_scrot
It now works on Colab. But the buttons color should have been black. But this is aesthetic issue and shouldn't block the work for writing a *working* advanced tutorial. @tpike3 this is ready to merge.

rht:
And the button symbols being very faded gray is specific to Colab. They are black on local Jupyter.

tpike3 (approved):
LGTM

Corvince:
Could you explain the reasoning for using a thread here? I didn't took a deep dive into your code or solara yet so I might be missing something, but I don't see any benefit of using a thread.

rht:
Threading is the only way (in the Python side of the notebook) to pause/stop a live loop of the steps. But it looks like threads in Colab are messy, not exactly how it behaves with vanilla Jupyter env. This PR instead does the loop from the JS that periodically triggers `model.step()`.

Corvince:
> Threading is the only way (in the Python side of the notebook) to pause/stop a live loop of the steps. But it looks like threads in Colab are messy, not exactly how it behaves with vanilla Jupyter env. This PR instead does the loop from the JS that periodically triggers `model.step()`.

Ah I see because the render loop runs on the same process of course. Have you tried the use thread   hook?
https://solara.dev/api/use_thread

rht:
I tested that `use_thread`. It works in `solara run`, but not local Jupyter env, nor Colab env. cc: @maartenbreddels.
Code can be found at https://github.com/rht/mesa-examples/tree/solara_use_thread.

rht:
By "not working", I meant that there was no visible sign that the plots are being updated at all.

rht:
The traceback
```
  File "/home/rht/code/venv/lib/python3.10/site-packages/reacton/ipyvue.py", line 23, in handler
    callback_ref.current(*args)
  File "/home/rht/code/venv/lib/python3.10/site-packages/solara/components/button.py", line 76, in <lambda>
    ipyvue.use_event(btn, click_event, lambda *_ignore: on_click and on_click())
  File "/home/rht/github/projectmesa/mesa-examples/experimental/jupyter_viz.py", line 201, in <lambda>
    # version is definite a much better implementation,
  File "/home/rht/code/venv/lib/python3.10/site-packages/reacton/core.py", line 1298, in set_
    self._possible_rerender()
  File "/home/rht/code/venv/lib/python3.10/site-packages/reacton/core.py", line 1335, in _possible_rerender
    self.render(self.element, self.container)
  File "/home/rht/code/venv/lib/python3.10/site-packages/reacton/core.py", line 1407, in render
    raise RuntimeError(msg)
RuntimeError: Too many renders triggered, your render loop does not stop
Last reason: Reason: state changed with key 0
Value changed from         Gini
0     0.0000
1     0.4736
2     0.5160
3     0.5632
4     0.5216
...      ...
3708  0.6792
3709  0.6984
3710  0.6528
3711  0.6448
3712  0.6400

[3713 rows x 1 columns] to         Gini
0     0.0000
1     0.4736
2     0.5160
3     0.5632
4     0.5216
...      ...
3709  0.6984
3710  0.6528
3711  0.6448
3712  0.6400
3713  0.6904

(repeated many times)

No such comm: 56d338386a9742e6a361ed631d21200b
```

I have added `time.sleep(0.3)` in my loop. It did slow down the simulation tick, but still the same error message.

Corvince:
I see. What I think is happening here is that the model runs too fast. use-thread is meant for long running tasks, that only triggers a change after some delay. But in this case the model state changes very quickly and every model change triggers a new render that probably doesn't even complete before the next render is triggered. The separate thread doesn't block the main thread but the changes happen too quickly. 

I think triggering a new step only every n milliseconds (like we currently do in ModularServer with the fps control) is a simpler solution with less room for getting out of sync.

Corvince:
Well probably my comment is only half true. It could also be that you continuously trigger state changes. Your threading function looks a bit strange, could it be that it never completes the thread?

In any case I think a non-threading solution would save us from a lot of headache

PR #50, opened 2023-07-06
trading Sugarscape: Add solara visualization

rht:


examples/sugarscape_g1mt/app.py
```diff
@@ -0,0 +1,61 @@
+import numpy as np
+import solara
+from matplotlib.figure import Figure
+from mesa_models.experimental import JupyterViz
+from sugarscape_g1mt.model import SugarscapeG1mt
+from sugarscape_g1mt.resource_agents import Sugar
+from sugarscape_g1mt.trader_agents import Trader
+
+
+def space_drawer(viz):
```
rht, line 10: @tpike3 what do you think of this custom space drawer? I use `imshow` because doing `ax.scatter()` on 2x50x50 resources is very slow. This would have been much harder to write with old Mesa frontend framework, which needs to be written in Python and JS.

examples/sugarscape_g1mt/app.py
```diff
@@ -0,0 +1,61 @@
...
+    "height": 50,
+}
+
+page = JupyterViz(
+    SugarscapeG1mt,
+    model_params,
+    measures=["Trader", "Price"],
+    name="Sugarscape {G1, M, T}",
+    space_drawer=space_drawer,
+    play_interval=1500,
```
rht, line 59: This has to be increased from the default 400 ms, because otherwise there is a "traffic jam" of the steps, which makes it impossible to pause the simulation immediately. The correct implementation should have been to use delay between steps instead of interval. But this is not doable via `ipywidgets.Play`. This requires the threading version instead. Solara dev is currently working to make `use_threading` work.

examples/sugarscape_g1mt/app.py
tpike3: That is awesome! Did you have to worry about the origin with imshow? If I remember correctly it is in the upper left vs lower left, as the mesa default?

rht: Did you have to invert when using `imshow` in the complexity tutorial? The i, j convention of `imshow` here is the same as the `scatter`.

tpike3: > Did you have to invert when using `imshow` in the complexity tutorial? The i, j convention of `imshow` here is the same as the `scatter`.

Yes for the tutorial I had to do ---- plt.imshow(resource_distribution, origin="lower")

rht: Should I do that too in this PR? It's hard to check after each steps if the resource being eaten corresponds to the agents' location.

tpike3 (approved):
This is super awesome!

examples/sugarscape_g1mt/app.py
```diff
@@ -0,0 +1,63 @@
...
+
+def space_drawer(viz):
+    def portray(g):
+        layers = {
+            "sugar": [[np.nan for j in range(g.height)] for i in range(g.width)],
+            "spice": [[np.nan for j in range(g.height)] for i in range(g.width)],
+            "trader": {"x": [], "y": [], "c": "tab:red", "marker": "o", "s": 10},
+        }
+
+        for i in range(g.width):
```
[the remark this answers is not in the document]
rht: I forgot about it. `coord_iter` should reduce the number of nesting.

PR #51, opened 2023-07-07
solara trading sugarscape: Use coord_iter

rht:


rht:
This is preferable because `model.grid._grid` is an internal attribute, and its direct access should be discouraged.

rht:
2023-07-07-101500_1210x568_scrot
More realistic example of what it looks like.

examples/sugarscape_g1mt/app.py
```diff
@@ -15,20 +15,19 @@ def portray(g):
...
 
-        for i in range(g.width):
-            for j in range(g.height):
-                content = g._grid[i][j]
-                for agent in content:
-                    if isinstance(agent, Trader):
-                        layers["trader"]["x"].append(i)
-                        layers["trader"]["y"].append(j)
+        # TODO update to Mesa 2.0 API
+        for content, i, j in g.coord_iter():
```
tpike3, line 19: Thanks @Corvince !

tpike3 (approved):
LGTM

PR #56, opened 2023-08-17
solara: Add virus_on_network network example

rht:
Waiting until https://github.com/projectmesa/mesa/pull/1767 is merged and we have a release.

### DIFF
--- a/examples/virus_on_network/app.py
+++ b/examples/virus_on_network/app.py
@@ -1,0 +1,135 @@
+import math
+
+import solara
+from matplotlib.figure import Figure
+from matplotlib.ticker import MaxNLocator
+from mesa.experimental import JupyterViz, make_text
+from virus_on_network.model import State, VirusOnNetwork, number_infected
+
+
+def agent_portrayal(graph):
+    def get_agent(node):
+        return graph.nodes[node]["agent"][0]
+
+    edge_width = []
+    edge_color = []
+    for u, v in graph.edges():
+        agent1 = get_agent(u)
+        agent2 = get_agent(v)
+        w = 2
+        ec = "#e8e8e8"
+        if State.RESISTANT in (agent1.state, agent2.state):
+            w = 3
+            ec = "black"
+        edge_width.append(w)
+        edge_color.append(ec)
+    node_color_dict = {
+        State.INFECTED: "tab:red",
+        State.SUSCEPTIBLE: "tab:green",
+        State.RESISTANT: "tab:gray",
+    }
+    node_color = [node_color_dict[get_agent(node).state] for node in graph.nodes()]
+    return {
+        "width": edge_width,
+        "edge_color": edge_color,
+        "node_color": node_color,
+    }
+
+
+def get_resistant_susceptible_ratio(model):
+    ratio = model.resistant_susceptible_ratio()
+    ratio_text = r"$\infty$" if ratio is math.inf else f"{ratio:.2f}"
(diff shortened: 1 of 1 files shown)


examples/virus_on_network/app.py
```diff
@@ -0,0 +1,136 @@
...
+    infected_text = str(number_infected(model))
+
+    return "Resistant/Susceptible Ratio: {}<br>Infected Remaining: {}".format(
+        ratio_text, infected_text
+    )
+
+
+def make_plot(viz):
+    # This is for the case when we want to plot multiple measures in 1 figure.
+    # We could incorporate this into core Mesa.
```
rht, line 53: Should this be merged to core Mesa for convenience?

rht:
This is ready to merge.

examples/virus_on_network/app.py
```diff
@@ -0,0 +1,134 @@
...
+from matplotlib.ticker import MaxNLocator
+from mesa.experimental import JupyterViz, make_text
+from virus_on_network.model import State, VirusOnNetwork, number_infected
+
+
+def agent_portrayal(graph):
+    def get_agent(node):
+        return graph.nodes[node]["agent"][0]
+
+    width = []  # edge width
```
Corvince, line 14: Maybe just call it edge_width in line with edge_color, so it doesn't need the comment?
```suggestion
    edge_width = []
```

examples/virus_on_network/app.py
```diff
@@ -0,0 +1,134 @@
...
+    edge_color = []
+    for u, v in graph.edges():
```
Corvince, line 16: u, v, w and ec are quite not-descriptive. Maybe call it edge_width, edge_color here and above edge_colors*s* and edge_width*s* ?

examples/virus_on_network/app.py
```diff
@@ -0,0 +1,134 @@
...
+def make_plot(viz):
+    # This is for the case when we want to plot multiple measures in 1 figure.
+    # We could incorporate this into core Mesa.
+    fig = Figure()
+    ax = fig.subplots()
+    measures = ["Infected", "Susceptible", "Resistant"]
+    colors = ["tab:red", "tab:green", "tab:gray"]
+    for i, m in enumerate(measures):
+        color = colors[i]
+        ax.plot(viz.df.loc[:, m], label=m, color=color)
```
Corvince, line 58: This needs to be changed to be dependent on model, not viz

Corvince (commented):
You need to change from viz to model, otherwise just some minor suggestions.

examples/virus_on_network/app.py
rht: Fixed.

examples/virus_on_network/app.py
rht: Fixed.

examples/virus_on_network/app.py
rht: `edge_color` in NetworkX is a list of edge colors. As much as this is confusing, I have to stay consistent with NetworkX. Ditto with the plural for `edge_width`.

`u`, `v` should be fine for undirected graph, just like `i`, `j` for 2D grid.

rht: See `edge_color` in https://networkx.org/documentation/stable/reference/generated/networkx.drawing.nx_pylab.draw_networkx.html, which may be a scalar or vector.

PR #65, opened 2023-11-01
Test examples

Corvince:
Added a test that initiales and runs each model for 10 steps. Currently only handles core examples, not geo examples. 

To make the test file run I had to modify some imports to be relative imports.

Otherwise this PR fixes 1 error and 1 warning:

In the epstein_civil_violence model the `count_cops` method was missing

The hexgrid examples used `HexGrid`, which is deprecated and should be changed to `HexGridSingle`.

### DIFF
--- a/examples/bank_reserves/bank_reserves/agents.py
+++ b/examples/bank_reserves/bank_reserves/agents.py
@@ -11,6 +11,7 @@ Author of NetLogo code:
 """
 
 import mesa
+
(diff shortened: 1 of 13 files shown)


rht:
Ruff check didn't pass.

rht (approved):

EwoutH:
And now somehow we need to link this to the main Mesa CI

PR #68, opened 2023-11-08
Boid: Rename velocity to direction_vector

EwoutH:
A velocity has a magnitude and a direction. Since the velocity was modelled as a unit vector, the magnitude is constant, and it only represents the direction, not the velocity (which is magnitude + direction). Therefor, rename `velocity` to `direction_vector`.

In this model, the magnitude is represented by `speed`.

Also removed a redundant division.

### DIFF
--- a/examples/boid_flockers/boid_flockers/boid.py
+++ b/examples/boid_flockers/boid_flockers/boid.py
@@ -99,7 +99,7 @@ class Boid(mesa.Agent):
             self.cohere(neighbors) * self.cohere_factor
             + self.separate(neighbors) * self.separate_factor
             + self.match_heading(neighbors) * self.match_factor
-        ) / 2
+        )
         self.direction_vector /= np.linalg.norm(self.direction_vector)
         new_pos = self.pos + self.direction_vector * self.speed
         self.model.space.move_agent(self, new_pos)
--- a/examples/boid_flockers/boid_flockers/boid.py
+++ b/examples/boid_flockers/boid_flockers/boid.py
@@ -12,7 +12,7 @@ class Boid(mesa.Agent):
         - Alignment: try to fly in the same direction as the neighbors.
 
     Boids have a vision that defines the radius in which they look for their
-    neighbors to flock with. Their speed (a scalar) and velocity (a vector)
+    neighbors to flock with. Their speed (a scalar) and direction (a vector)
     define their movement. Separation is their desired minimum distance from
     any other Boid.
     """
@@ -23,7 +23,7 @@ class Boid(mesa.Agent):
         model,
         pos,
         speed,
-        velocity,
+        direction_vector,
         vision,
         separation,
         cohere=0.025,
@@ -37,6 +37,7 @@ class Boid(mesa.Agent):
             unique_id: Unique agent identifyer.
             pos: Starting position
             speed: Distance to move per step.
+            direction_vector: numpy vector for the Boid's direction of movement. It is a unit vector (a vector with length 1) that determines the direction of the Boid's movement.
             heading: numpy vector for the Boid's direction of movement.
             vision: Radius to look around for nearby Boids.
             separation: Minimum distance to maintain from other Boids.
@@ -47,7 +48,7 @@ class Boid(mesa.Agent):
         super().__init__(unique_id, model)
         self.pos = np.array(pos)
         self.speed = speed
-        self.velocity = velocity
+        self.direction_vector = direction_vector
         self.vision = vision
         self.separation = separation
         self.cohere_factor = cohere
@@ -84,7 +85,7 @@ class Boid(mesa.Agent):
         match_vector = np.zeros(2)
         if neighbors:
             for neighbor in neighbors:
-                match_vector += neighbor.velocity
+                match_vector += neighbor.direction_vector
             match_vector /= len(neighbors)
         return match_vector
 
@@ -94,11 +95,11 @@ class Boid(mesa.Agent):
         """
 
         neighbors = self.model.space.get_neighbors(self.pos, self.vision, False)
-        self.velocity += (
+        self.direction_vector += (
             self.cohere(neighbors) * self.cohere_factor
             + self.separate(neighbors) * self.separate_factor
             + self.match_heading(neighbors) * self.match_factor
         ) / 2
-        self.velocity /= np.linalg.norm(self.velocity)
-        new_pos = self.pos + self.velocity * self.speed
+        self.direction_vector /= np.linalg.norm(self.direction_vector)
+        new_pos = self.pos + self.direction_vector * self.speed
         self.model.space.move_agent(self, new_pos)
--- a/examples/boid_flockers/boid_flockers/model.py
+++ b/examples/boid_flockers/boid_flockers/model.py
@@ -58,13 +58,13 @@ class BoidFlockers(mesa.Model):
             x = self.random.random() * self.space.x_max
             y = self.random.random() * self.space.y_max
             pos = np.array((x, y))
-            velocity = np.random.random(2) * 2 - 1
+            direction_vector = np.random.random(2) * 2 - 1
             boid = Boid(
                 i,
                 self,
                 pos,
                 self.speed,
-                velocity,
+                direction_vector,
                 self.vision,
                 self.separation,
                 **self.factors,


examples/boid_flockers/boid_flockers/boid.py
```diff
@@ -94,11 +95,11 @@ def step(self):
         """
 
         neighbors = self.model.space.get_neighbors(self.pos, self.vision, False)
-        self.velocity += (
+        self.direction_vector += (
             self.cohere(neighbors) * self.cohere_factor
             + self.separate(neighbors) * self.separate_factor
             + self.match_heading(neighbors) * self.match_factor
-        ) / 2
```
rht, line 101: The `/ 2` affects the result. This is `self.velocity +=`, not `self.velocity =`.

tpike3 (approved):
LGTM

rht:
I agree with the rename, but the removal of the division by 2 has a problem. To rephrase what I said, adding a unit vector with another vector with a different magnitude (with or without division by 2) and then to normalize it to unit vector again will result with a unit vector with a different direction.

examples/boid_flockers/boid_flockers/boid.py
EwoutH: Good catch, you're totally right. I will reply in the main thread more extensively.

EwoutH:
You're right on the updating and not overwriting, I read over that.

However:
- is updating the desired behavior? Should they update their direction every step (like now) or should they just create a new direction anyways?
- Where does the / 2 come from? Is it correcting for that there are three factors (shouldn't it be `/ 3`). Or is it some weight? In that case, `* 0.5` is more logical, and it should be maybe an input variable (like `update_direction_weight`).

The `/ 2` feels like a magic number. Do we know where it came from and why?

rht:
> is updating the desired behavior?

Updating makes more sense, because it implements an acceleration. There should be an inertia in the boid movement.

> Where does the / 2 come from?

Not sure. https://www.red3d.com/cwr/boids/ has the original implementation
```lisp
(defmethod (:NAVIGATE BOID-BRAIN) ()
  (vlet* ((avoid-obstacles   (send self :environmental-obstacle-avoidance))
	  (avoid-flockmates  (send self :flockmate-avoidance))
	  (velocity-matching (send self :velocity-matching))
	  (centering-urge    (send self :neighborhood-centering))
	  (migratory-urge    (send self :migratory-urge))
	  (course-leveling   (send self :course-leveling))
	  (course-damping    (send self :course-damping)))
    ;;
    ;; Available-acceleration should probably be 1.0, but I've set it a little higher to
    ;; avoid have to readjust the weighting factors for all of the acceleration requests.
    ;;
    (vlet* ((composite-acceleration (prioritized-acceleration-allocation 1.3 ;; 1.0
									 avoid-obstacles
									 avoid-flockmates
									 velocity-matching
									 centering-urge
									 migratory-urge
									 course-leveling
									 course-damping))
	    (weights                (values-list (send self :steering-weights)))
	    (desired-acceleration   (componentwise-product weights composite-acceleration)))
      '(when (eql self (default-boid))
	 (format w " n"))
      (values-list desired-acceleration))))

(defun PRIORITIZED-ACCELERATION-ALLOCATION (available-acceleration &rest vectors)
  (loop with allocated-acceleration = 0
	with (tx ty tz) = '(0 0 0)
	for v in vectors
	for (vx vy vz) = (or v '(0 0 0))
	for mag-v = (magnitude v)
	for scaler = (if (> (+ allocated-acceleration mag-v) available-acceleration)
			 (// (- available-acceleration allocated-acceleration) mag-v)
			 1)
	do (incf tx (* vx scaler))
	   (incf ty (* vy scaler))
	   (incf tz (* vz scaler))
	   (incf allocated-acceleration mag-v)
	   ;; (format tow "~&  (~d ~D ~d), ~d, ~d" tx ty tz allocated-acceleration mag-v)
	until (> allocated-acceleration available-acceleration)
	finally
	  ;; (format t "~&  mag= ~d" (magnitude-xyz tx ty tz))
	  ;; (format w "~&~D" allocated-acceleration)
	  (return (values tx ty tz))))
```
`update_direction_weight` makes sense to me. But it seems to differ from `available-acceleration` in that code.

EwoutH:
Interesting. If we’re actually talking about acceleration, I find it incredibly weird that it get’s normalized to a unit vector each step.

Can we find the original model conceptualization or formalization?

rht:
Superseded by #104, except for the discussion.

@coderbeta1 if you want to read the original implementation of Boid Flockers, the thread above has a link and relevant code snippet to it. We might have deviated from the original implementation by not implementing acceleration.

EwoutH:
I think the most important distinction that needs to be made is if you want to use a fixed coordinate system, of if you want to use the boid itself as the frame of reference.

Both have advantages. I'm nudging towards the later, because there is a good case to be made that boids have more control for changing their heading than for changing their speed.

coderbeta1:
>a good case to be made that boids have more control for changing their direction (heading) than for changing their velocity (speed)

True that. This just means that with the proposed implementation we increase the `max_force`. This would mean that the boids are more capable of changing direction (since top speed is limited).

If we talk about intuition I think it completely depends on the person. 

For a third person viewer the stationary coordinate system is more intuitive. This is because 
* it is easy to see the cluster and imagine things that way.
* Concept of Centre of Mass (COM) is more intuitive when we see the whole picture. 

On the other hand,
* Using the frame of reference of the boid will let us apply changes easily but it is a little more difficult to imagine change in direction towards COM. 
* Using an  accelerating frame of reference and imaging a pseudo force is more convenient

Calculating the forces is easy to implement either way. I feel that the way the [article](https://betterprogramming.pub/boids-simulating-birds-flock-behavior-in-python-9fff99375118) is implemented is quite intuitive when is comes to acceleration.

rht:
I agree with proceeding with the implementation with acceleration and center of mass. Let's proceed forward. The [citation](https://cs.stanford.edu/people/eroberts/courses/soco/projects/2008-09/modeling-natural-systems/boids.html) (pseudocode at https://vergenet.net/~conrad/boids/pseudocode.html) at the README.md also seems to suggest a form of acceleration: 

This is the original PR that implemented the example: https://github.com/projectmesa/mesa/pull/136, which has no information about the source of the model.

coderbeta1:
ok so I tried to follow the pseudocode as mentioned by @rht . After spending a lot of time reimplementing it, what I noticed is that the [original implementation](https://github.com/projectmesa/mesa-examples/blob/main/examples/boid_flockers/boid_flockers/model.py) is the same as https://vergenet.net/~conrad/boids/pseudocode.html . Its just that the code implements it in a way that is not easy to understand (since there are no comments in the logic part).

* The part of `centre of mass` is actually implemented very cleverly already. (It is just implemented in a different frame of reference - with boid as the origin) 
* In the [citation](https://cs.stanford.edu/people/eroberts/courses/soco/projects/2008-09/modeling-natural-systems/boids.html) they implement static values of scaling factor for cohesion, separation and alignment. But with the current implementation it is easy to fine tune and check the importance of the factors

One thing that I feel should be changed is the scheduling.
* For this case specifically I think Simultaneous Activation is more logical than random. I tried to implement it and the result seems to be better with simultaneous activation. It just feels more smooth and predictable.
* If we consider a cluster of boids with random activation and a cluster with simultaneous, it just seems logical to go with simultaneous. This is because if we take an example of (lets say) 3 boids and do random activation, the order of steps will make a huge difference (and in some cases compound the result)

rht:
@coderbeta1 you are doing a scholarly work here! We should update [Agents.jl](https://github.com/JuliaDynamics/Agents.jl) about your findings once the dust has settled, as they use Boid for https://github.com/JuliaDynamics/ABMFrameworksComparison.

I have a question regarding with the acceleration: the v1, v2, v3, ... in https://vergenet.net/~conrad/boids/pseudocode.html are technically the `forces / mass * deltat`. But in the current mesa-examples implementation, the speed is held constant, and only the direction changes. Does the acceleration part differ from the citation?

I agree with the SimultaneousActivation being more logical. Eliminating first mover advantage and order hysteresis makes sense.

EwoutH:
Awesome work so far!

> For this case specifically I think Simultaneous Activation is more logical than random. 

Agreed! It's also faster and less noisy, and all agents act in continuous time independent from each other, so it's obviously the way to go. RandomActivation also allows an agent acting two times in a row, which from a reaction-time standpoint is not the proper way to model it.

rht:
If you look at Reynold's original implementation in Lisp, https://www.red3d.com/cwr/code/boids.lisp, it has `min-speed`, `max-speed`, and `max-accel-relative`. But it also has `acceleration-smoothing`, a gravitational field, etc. It's too detailed, but maybe we could have a folder boid_flockers_full_version later on.

If you look at the pseudocode in https://vergenet.net/~conrad/boids/pseudocode.html, it does have a speed limit. Maybe just limiting the speed would be sufficient, without having to set an upper bound to the acceleration?

coderbeta1:
Apologies for the late reply, got some important work on my side.

> If you look at the pseudocode in https://vergenet.net/~conrad/boids/pseudocode.html, it does have a speed limit. Maybe just limiting the speed would be sufficient, without having to set an upper bound to the acceleration?

The speed limit does help, but it does not work if acceleration does not have a limit.
* In general I noticed that with max acceleration < max speed I got best results
* Changing the vision radius of boid has a massive difference on the solution. High radius gives best results (it does expectedly slow down the simulation though)

> It's too detailed, but maybe we could have a folder boid_flockers_full_version later on.

* Seems like a great idea.
* I think for now I'll just create a PR for `random sequence` to `simultaneous` and the `max_speed` + `max_acceleration` change

PR #70, opened 2023-11-26
Sugarscape {G1,{M,T}}: Add tests

rht:

PR #71, opened 2023-11-27
refactor: Simplify Sugarscape resource agents

rht:
This combines the Sugar and Spice resources into 1 resource class.
Rational:
- More concise code, which then requires less explanation
- Scales better with increasing number of resource types: see https://github.com/projectmesa/mesa/issues/1889#issuecomment-1827777988.

### DIFF
--- a/examples/sugarscape_g1mt/sugarscape_g1mt/trader_agents.py
+++ b/examples/sugarscape_g1mt/sugarscape_g1mt/trader_agents.py
@@ -2,7 +2,7 @@ import math
 
 import mesa
 
-from .resource_agents import Spice, Sugar
+from .resource_agents import Resource
 
 
 # Helper function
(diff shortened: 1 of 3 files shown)


rht:
I tested with the tests in #70, and confirmed that the resulting timeseries is unchanged.

examples/sugarscape_g1mt/sugarscape_g1mt/trader_agents.py
```diff
@@ -50,48 +50,13 @@ def __init__(
...
-        used in self.get_sugar_amount()
-        """
-
+    def get_resource(self, pos):
         this_cell = self.model.grid.get_cell_list_contents(pos)
         for agent in this_cell:
-            if type(agent) is Sugar:
+            if type(agent) is Resource:
                 return agent
         return None
```
tpike3, line 58: I don't think this ever called. Under this set up there is a resource agent in every cell even if they have 0 sugar and 0 spice. If it was called line 260 and 261 would error out with `None + int` Correct?

rht: I agree that the `return None` not being called because every cells has a `Resource` agent is rather an implicit consequence. A type checker would have said that `self.get_resource(pos).sugar_amount` does not hold for all cases.

It should be replaced with a patch/cell object soon, so that the return value of `get_resource` is `Resource`, not `Resource | None`.

rht: I raised an `Exception` in `get_resource` when no `Resource` agent is found. This should pass Mypy check.

PR #72, opened 2023-11-30
fix: Get all traders in neighborhood instead of 1 trader each cell

rht:
TODO:
- [ ] test for the changes when the density of the trader agents is higher


EwoutH:
Can't you use `get_neighbors()` here?

PR #79, opened 2024-01-07
Boltzmann wealth model: Use the new model.agents API

rht:


rht:
This is ready for review.

examples/boltzmann_wealth_model_experimental/model.py
```diff
@@ -54,23 +51,25 @@ def __init__(self, unique_id, model):
...
         if len(cellmates) > 0:
             other = self.random.choice(cellmates)
             other.wealth += 1
             self.wealth -= 1
 
     def step(self):
         self.move()
-        if self.wealth > 0:
-            self.give_money()
+        self.give_money()
```
EwoutH, line 75: I liked the old check here better, it made it explicit that only money was given when there was any to give.

EwoutH (approved):
Thanks! One thing I would like to have changed, otherwise looks good!

examples/boltzmann_wealth_model_experimental/model.py
rht: What about `maybe_give_money`?

EwoutH: Yeah renaming the function would also work (`consider_giving_money` maybe), but for explicitness I liked the old structure better.

rht: My reasoning is different. The check gives a constraint to the `give_money` method that prevents negative wealth.

EwoutH: @Corvince @tpike3 @quaquel what do you think?

rht: `consider_giving_money` means making a decision to give money, not necessarily executing it. For `maybe_give_money`, it's either or, and the `maybe_*` pattern has already been used in Sugarscape examples.

Corvince: I also think the older version was better, because more explicit. "Maybe" could also mean it's a random decision. 

If worried about potentially leading to negative money, because the function might be called from somewhere else I see no problem duplicating the check inside the function or even raise an exception - if you have no money but try to give some that should be an error.

But that brings me to the idea of calling it try_give_money which might work

rht: The name choice actually comes from Haskell's [Maybe monad](https://en.wikibooks.org/wiki/Haskell/Understanding_monads/Maybe) (this [link](https://www.futurelearn.com/info/courses/functional-programming-haskell/0/steps/27247) has a simpler example).

> because the function might be called from somewhere else I see no problem duplicating the check inside the function or even raise an exception

The problem is that this requires discipline to add the wealth check whenever the method is called. Might as well to sew it in the method so that it is more portable.

Corvince: Huh? If the naming is based on the maybe monad then it should have another name, no? A maybe monad either returns a value or no value. But the function never returns a value, so I don't understand how that should relate?

rht: Of course it doesn't mean it is a monad, just that it is not probabilistic. It is a method that either does something or nothing. If you are concerned about the connotation with the monad, then there is this usage in [Golang](https://github.com/search?q=repo%3Agolang%2Fgo%20maybe&type=code).

Corvince: I am just curious why you said the name choice is based on Haskells Maybe monad. If it's just about maybe doing something, then the dictionary definition of maybe would suffice. What am I missing?

rht: I have already said this
> If you are concerned about the connotation with the monad, then there is this usage in [Golang](https://github.com/search?q=repo%3Agolang%2Fgo%20maybe&type=code).

It's just my first encounter on the usage of the term "maybe" in naming in the context of programming.

Corvince: And I have already said that I am not concerned but just curious 😅

I have no idea what you want to tell me, but we should probably just leave this conversation. It's not important.

EwoutH: > I also think the older version was better, because more explicit.

I agree. It's more explicit and more readable. Let's leave it the old way.

rht: Reverted.

tpike3:
The tests are failing is this because the tests are now wrong? (Apologies for the delay)

rht:
Fixed in https://github.com/projectmesa/mesa/pull/2050.

EwoutH:
Looks good to go!

quaquel:
I agree, this can be merged.

rht:
If this is merged, anyone trying to run the example from the latest stable release will encounter an error.

PR #80, opened 2024-01-08
Schelling: Use new API and simplify

rht:
The goal is to reduce unnecessary boilerplate code so as to approach Agents.jl's [Schelling](https://github.com/JuliaDynamics/ABMFrameworksComparison/blob/main/Schelling/Agents/Schelling.jl) example in terms of conciseness.


tpike3 (approved):
LGTM

PR #82, opened 2024-01-11
ci: Add examples testing & Move Ruff and Black to pre-commit

rht:


Corvince:
Looks much cleaner now! 

But somehow the action added in this PR isn't run. I have no idea why. There is also a TODO added, do you want to address this in this PR or is it ready for review? (pre-commit is failing)

rht:
I have made sure the GH Actions is running now. However, the errors that have shown up seem to be real bugs in the current 2.2.0 release.

rht:
> There is also a TODO added, do you want to address this in this PR or is it ready for review? (pre-commit is failing)

I will defer the TODO (linting the .ipynb) to later.

Corvince:
> I have made sure the GH Actions is running now. However, the errors that have shown up seem to be real bugs in the current 2.2.0 release.

So the PR already proofed itself useful! It seems like 1 error might be a true bug and the other 2 (which are the same) are involuntarily added breaking changes. But I think we can merge this, because fixing those is outside the scope of this PR. Do you agree?

Corvince (approved):

rht:
I agree with merging this PR now, so as to show reminder of the bugs, until they are fixed.

rht:
It probably makes more sense if the tests in the GH Actions in this PR are run in projectmesa/mesa instead of here? The CI tests in this repo are only run when the examples are modified.

Corvince:
It makes sense in both places. Here if new/modified examples fail and in mesa if it breaks the examples. Would have catched the current bugs early

PR #83, opened 2024-01-11
Fix examples for Mesa 2.2.0 API change

rht:


rht:
Ready for review. This PR fixes all the issues except for the weakref stuff, which needs to happen in projectmesa/mesa itself.

tpike3 (approved):
Thanks for doing this @rht!

PR #89, opened 2024-01-22
Fix two more cases of a missing super().__init__() in the model init

EwoutH:
Fixes two more cases of a missing super().__init__() in the model init. Should resolve two of the errors in https://github.com/projectmesa/mesa/pull/1942.


EwoutH:
Please squash on merging

rht (approved):

PR #91, opened 2024-01-24
Revert "fix: Use new API to loop over agents_by_type"

rht:
This reverts commit ef2b5e735140385973cdfa7894f45253bd48cd62.


EwoutH (approved):
The current dict is structured `{unique_id: Agent}` right? In that case, looks good.

PR #92, opened 2024-01-24
fix: sugarscape G1MT: Remove redundant schedule steps increment

rht:


rht:
cc: @tpike3

EwoutH:
Could you add a brief description what changed exactly that this isn't needed anymore? That helps to place it in context.

rht:
This is a long-standing bug. The `RandomActivationByType` already has https://github.com/projectmesa/mesa/blob/c0c8479e4d1f29c20ee2a32caeb6a3eaa1682041/mesa/time.py#L363.

rht:
Wait, never mind. self.schedule.steps() never gets called. This is probably not the right fix. @tpike3 may know the problem better.

PR #93, opened 2024-01-24
fix: Ensure schedule.steps is always incremented before data collection

rht:
This replaces #92.

### DIFF
--- a/examples/sugarscape_g1mt/sugarscape_g1mt/model.py
+++ b/examples/sugarscape_g1mt/sugarscape_g1mt/model.py
@@ -168,6 +168,9 @@ class SugarscapeG1mt(mesa.Model):
 
         if not self.enable_trade:
             # If trade is not enabled, return early
+            self.schedule.steps += (
+                1  # important for data collector to track number of steps
+            )
             self.datacollector.collect(self)
             return
 


EwoutH:
Could you describe the problem and why this solution fixed it, for clarity?

rht:
The problem is when trade is disabled, the steps never get incremented because the code exits early in this code branch. Needs a manual `self.schedule.steps += 1`, because there is no `self.schedule.step()` call. The trading Sugarscape needs custom scheduling that is beyond staged activation, and is more easily expressed by the AgentSet API.

EwoutH:
Would this work? If so, I think it's more elegant:

```Python
def step(self):
    """
    Unique step function that does staged activation of sugar and spice
    and then randomly activates traders
    """
    # Step Resource agents
    for resource in self.schedule.agents_by_type[Resource]:
        resource.step()

    # Step trader agents
    trader_shuffle = self.randomize_traders()
    for agent in trader_shuffle:
        agent.prices = []
        agent.trade_partners = []
        agent.move()
        agent.eat()
        agent.maybe_die()

    if self.enable_trade:
        # If trade is enabled, proceed with trading
        for agent in trader_shuffle:
            agent.trade_with_neighbors()

    # Update step count and collect data
    self.schedule.steps += 1
```

rht:
No it doesn't, because there are more lines of code beyond `self.datacollector.collect(self)` when trade is enabled: https://github.com/projectmesa/mesa-examples/blob/cf38477725771d4658ea6dcf393766f1feefcc88/examples/sugarscape_g1mt/sugarscape_g1mt/model.py#L188-L205.

rht:
We can optimize the code clarity later. This PR's scope is to fix the test issue with a very localized patch.

EwoutH (approved):
Indeed, this PR fixes the issue. We can refactor later.

PR #95, opened 2024-01-24
sugarscape_g1mt: Remove dependence on model.schedule.steps

rht:
Depends on https://github.com/projectmesa/mesa/pull/1942 to be merged.


Corvince:
The necessity to modify an internal variable hints at more work to be done on the mesa side

EwoutH:
We probably want to reserve `Model.steps` and `Model.time` in some way.

PR #100, opened 2024-02-18
Improve Virus on Network documentation

coderbeta1:
Change 1: Make the ReadMe more descriptive

1. Added a more detailed explanation of the model
2. Added an extra method of running the model

Change 2: Doc Strings and Cleaning

Made small changes to make the code more readable and easy to understand at first glance

NOTE: In the model, after running, some of the nodes in the graph are isolated. In my opinion it should be connected (if it was intentional then fine but in case not it needs to be updated with a different graph creation technique)

### DIFF
--- a/examples/virus_on_network/app.py
+++ b/examples/virus_on_network/app.py
@@ -1,5 +1,4 @@
 import math
-
 import solara
 from matplotlib.figure import Figure
 from matplotlib.ticker import MaxNLocator
@@ -48,7 +47,6 @@ def get_resistant_susceptible_ratio(model):
 
 def make_plot(model):
     # This is for the case when we want to plot multiple measures in 1 figure.
-    # We could incorporate this into core Mesa.
     fig = Figure()
     ax = fig.subplots()
     measures = ["Infected", "Susceptible", "Resistant"]
--- a/examples/virus_on_network/app.py
+++ b/examples/virus_on_network/app.py
@@ -1,4 +1,5 @@
 import math
+
 import solara
 from matplotlib.figure import Figure
 from matplotlib.ticker import MaxNLocator
--- a/examples/virus_on_network/README.md
+++ b/examples/virus_on_network/README.md
@@ -32,7 +32,7 @@ To run the model interactively, run ``mesa runserver`` in this directory. e.g.
 
 Then open your browser to [http://127.0.0.1:8521/](http://127.0.0.1:8521/) and press Reset, then Run.
 
-or 
+or
 
 Directly run the file ``run.py`` in the terminal. e.g.
 
--- a/examples/virus_on_network/README.md
+++ b/examples/virus_on_network/README.md
@@ -2,7 +2,13 @@
 
 ## Summary
 
-This model is based on the NetLogo model "Virus on Network".
+This model is based on the NetLogo model "Virus on Network". It demonstrates the spread of a virus through a network and follows the SIR model, commonly seen in epidemiology.
+
+The SIR model is one of the simplest compartmental models, and many models are derivatives of this basic form. The model consists of three compartments:
+
+S: The number of susceptible individuals. When a susceptible and an infectious individual come into "infectious contact", the susceptible individual contracts the disease and transitions to the infectious compartment.
+I: The number of infectious individuals. These are individuals who have been infected and are capable of infecting susceptible individuals.
+R for the number of removed (and immune) or deceased individuals. These are individuals who have been infected and have either recovered from the disease and entered the removed compartment, or died. It is assumed that the number of deaths is negligible with respect to the total population. This compartment may also be called "recovered" or "resistant".
 
 For more information about this model, read the NetLogo's web page: http://ccl.northwestern.edu/netlogo/models/VirusonaNetwork.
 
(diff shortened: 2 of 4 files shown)


rht:
pre-commit checks failed: https://results.pre-commit.ci/run/github/42205414/1708283306.gaMzyn-PS4uAHxWwbDygnQ.

EwoutH:
pre-commit.ci autofix

examples/virus_on_network/app.py
```diff
@@ -1,5 +1,4 @@
 import math
-
 import solara
```
EwoutH, line 2: Our [ruff](https://github.com/astral-sh/ruff) formatter wants the empty line here: it separates the math import (a Python built-in) from the other imported packages.

EwoutH (commented):
Thanks for the PR, this is a valuable improvement to this example!

There is one minor issue with our formatter, once that's fixed it looks good to me to merge!

tpike3:
@coderbeta1 Welcome and thanks for the contribution! 

This is a super minor fix, if you would prefer we can do it on our end, just let us know. 

Thanks again

examples/virus_on_network/virus_on_network/__init__.py
EwoutH, line 1: I'm not sure, do we need (to keep) an empty init file in each folder? For tests, CI or else?

coderbeta1: I don't think it is practical to keep an empty file in the folder. It only leads to more confusion. Which is why I chose to remove it from the folder. For the sake of examples we should keep it as similar to [best practices](https://github.com/projectmesa/mesa/blob/main/docs/best-practices.md) as possible

For testing we still have the app.py file. Any test can be done in it. (Ideally the model can still be run without the app.py file. We could just run the server.py file with some small changes)

EwoutH: @rht do you know if we need it for something?

rht: It's fine to remove it in the folders that don't end with _experimental, due to the folder structure. Not having `__init__.py` will prevent relative import if you run a Python file from the same directory, which is the case for the `*_experimental` examples.

EwoutH:
Merged, thanks for your contribution!

examples/virus_on_network/virus_on_network/__init__.py
Corvince: @EwoutH  Thats not totally accurate. Yes, without a __init__.py in the same folder you can not do relative imports like 
`from .model imort VirusOnNetwork`
This means for virus_on_network without the file you can no longer import/execute the server.py file from the same folder where model.py is located.

But from the parent folder `import virus_on_network.server` still works for the `run.py` file!

On the other side for example the schelling_experimental `app.py` file imports via `from model import Schelling`. This *does* work even without an __init__.py file. But it is considered bad style because you don't know if `model` refers to a local module (like here) or a global package. 

So I think we should consider adding a __init__.py file to the best-practices and recommend relative imports. This way models are their own package in Python-speak.

rht: I am not @EwoutH. I'm not sure if you raised an issue for the `_experimental` folders or everything else. But given that

> But from the parent folder import virus_on_network.server still works for the run.py file!

And the subsequent paragraph. Does this mean you have no problem with "everything else"?

Corvince: > I am not @EwoutH.

I know that. I specifically addressed Ewout since he asked about the role of `__init__.py` files. I felt like your answer to that was incomplete and could be misinterpreted to mean:
1. We could remove most `__init__.py` files without any side effects
2. They are needed to import modules from the same folder. 

Both statements are incorrect and just wanted to make sure Ewout is aware of that. I explicitly did not want to start a discussion with, hence the @EwoutH tag.

But regarding your question, I don't have a problem with any of that. Otherwise I would have opened an issue. My last paragraph was just something to maybe keep in mind.

rht: > We could remove most __init__.py files without any side effects

There is [PEP 420](https://peps.python.org/pep-0420/) that introduced namespace packages, which removes the need for `__init__.py` in every package folders.

> They are needed to import modules from the same folder.

I tested this earlier and misread the result that the __init__.py is required for a relative import in the same folder. 
Still got `ImportError: attempted relative import with no known parent package`.

> I explicitly did not want to start a discussion with you, hence the @EwoutH tag.

It meant a discussion was started, but I was excluded because you unilaterally thought I was wrong again, but I have no available energy to further talk about this.

Corvince: I have no idea how you can interpret "Thats not totally accurate" as me thinking you are wrong. I think you are reading too much into what I wrote. My response was made as a clarification, not a correction of what you said. Sorry if that came across wrongly (but again, I wasn't addressing you in the first place. Although I do understand that this led to you feeling excluded. That was not my intention.)

rht: That clarifies things.

Issue #86, opened 2024-01-21
Improve boid flockers (Solara) visualisation with direction and colors


EwoutH:
The boid flockers example can be improved with (Solara) visualisation that includes the direction and colors for the boids. It would be nice to have an arrow-like shape to know the direction, and either random colors or to let them take the colors of a group/leader.

Here's an example, in matplotlib, from an old exam we used:

![Screenshot_351](https://github.com/projectmesa/mesa-examples/assets/15776622/19816294-d420-424b-be72-7f54e7870f2f)

```Python
# set the colormap we will use for plotting
bwr = plt.get_cmap("gist_rainbow")

def draw_boids(model):
    x_vals = []
    y_vals = []
    u_vals = []
    v_vals = []
    colors = []  # List to hold the color for each agent

    # Generate unique colors with saturation and transparency
    # Use a colormap with high saturation, such as "hsv"
    cmap = plt.get_cmap("hsv")
    n_agents = len(model.schedule.agents)
    color_indices = np.linspace(0, 1, n_agents)

    for i, boid in enumerate(model.schedule.agents):
        x, y = boid.pos
        u, v = boid.direction_vector
        x_vals.append(x)
        y_vals.append(y)
        u_vals.append(u)
        v_vals.append(v)
        
        # Assign a color with transparency; alpha value less than 1 introduces transparency
        colors.append(cmap(color_indices[i])[:3] + (0.6,))  # Change 0.6 to the desired transparency

    fig = plt.figure(figsize=(7, 7))
    ax = fig.add_subplot(111)

    # Use quiver with colors
    q = ax.quiver(x_vals, y_vals, u_vals, v_vals, color=colors, angles='xy', pivot='mid',
                  headaxislength=8, headlength=10, headwidth=7, scale=35)

    # The call to `embed_identity` is not changed.
    embed_identity(ax, student_name, student_number)

draw_boids(model)
```

EwoutH:
@quaquel you might also find this interesting for your exam model.

areyoush:
Hi @EwoutH,
May I work on this issue? if it’s still open. I have a good understanding of the problem.

(Also, I’ve recently implemented several contributor examples, including one based on the Deffuant–Weisbuch opinion dynamics model, which I think could be a good addition)

Please let me know if you’re okay with me proceeding.

areyoush:
[...]

PR #101, opened 2024-02-22
Improve boid flocker model and documentation

coderbeta1:
I have made the following changes:

1. ReadMe Update: Many links in the readme were outdated due to change in folder structure. I have fixed it to work with current folders
2. ReadMe Update: I have restructured the readme to make it similar to the readme of other examples. I have also added a link for additional information
3. General readability: I have made general improvements by making the code more easy to read for newcomers
4. Merge Agent and Model: I have deleted boid.py and merged it with model.py to follow [best practices](https://github.com/projectmesa/mesa/blob/main/docs/best-practices.md)
5. Renamed velocity: I have renamed the attribute "velocity" to "direction". Velocity is misleading since there is a speed attribute also and velocity is used to signify direction only in this case
6. Model Visualization: I made it so that if there are 0 or 1 neighbours then the agent is red, else green. This help in better visual
7. Sliders: I made it so that the user can now change the population and other parameters interactively

image

NOTE: I also tried to work on #86  for adding better visualization but I was facing some issues.

1. As for colour I made a simple colouring scheme such that if there are less than 2 neighbours the agent is red, else green (Implemented)
2.  When I tried to make changes to the JavaScript file to add arrowHeads my changes were not being reflected. It did reflect initially, but later on it just did not reflect even the smallest change. I even deleted the entire JS file but the code still worked (it should not work without the JS file). It seemed like the python file was no longer dependent on the JavaScript file. I tried reinstalling mesa and retried but it did not work.


examples/boid_flockers/Readme.md
```diff
@@ -1,34 +1,47 @@
...
+$ mesa runserver
+```
+
+or 
+
+Directly run the file ``app.py`` in the terminal. e.g.
+
 ```
-    $ python Flocker_Server.py
+    $ python app.py
```
rht, line 30: You meant run.py. app.py needs to be run with `solara run app.py`.
https://github.com/projectmesa/mesa-examples/blob/06c0015a8f0895092405996c6568d7ebf5741ef8/examples/boltzmann_wealth_model_experimental/Readme.md?plain=1#L21-L25

coderbeta1: You are right. I missed it while making the change. I have now corrected it and made the change from app.py to run.py

rht:
pre-commit.ci autofix

examples/boid_flockers/boid_flockers/server.py
```diff
@@ -5,19 +5,56 @@
 
 
 def boid_draw(agent):
-    return {"Shape": "circle", "r": 2, "Filled": "true", "Color": "Red"}
+    neighbors = agent.model.space.get_neighbors(agent.pos, agent.vision, False)
```
rht, line 8: This operation is very expensive for all agents, for each step. At which number of agents did you notice any performance degradation?

rht:
Except for the `boid_draw` performance measurement question, everything else LGTM.

examples/boid_flockers/boid_flockers/server.py
coderbeta1: For testing I set 20 FPS. It is fine until 200 agents (which I set as upper limit). After this it slows down kinda linearly until 500, after which it is unbearably slow.

To make the colour process faster I can add the colour logic in the agent class itself. This would make it faster since we anyways need to search neighbourhood there. I can make the change as per your recommendation.

examples/boid_flockers/boid_flockers/model.py
```diff
@@ -8,7 +8,107 @@
...
+                match_vector += neighbor.direction
+            match_vector /= len(neighbors)
+        return match_vector
+
+    def step(self):
+        """
+        Get the Boid's neighbors, compute the new vector, and move accordingly.
+        """
+
+        neighbors = self.model.space.get_neighbors(self.pos, self.vision, False)
```
EwoutH, line 103: If we make this self.neighbors we can also use if for the visualisation right?

coderbeta1: Yes that would work fine, but it could take too much memory with many agents. If the only need is to change colour, I think it would be more convenient to make a self.colour variable instead.

EwoutH: We could save the neighbours of the previous step.

I think in many systems that compute will be more of an bottleneck than memory

coderbeta1: I have made the changes as suggested

rht: > The model can now smoothly handle around 600 agents after which it slows down. After 1000 it is too slow and not practical

That's useful info to take a note on.

rht:
pre-commit.ci autofix

rht:
Merged, thank you.

examples/boid_flockers/boid_flockers/model.py
EwoutH: First it was around 200 right? If so, that's a great improvement! Thanks!

PR #102, opened 2024-02-23
Improve Caching and Replay Example

coderbeta1:
Changes Made:

1. Links Update: The previous link did not work. I updated it to make it work
2. Run Method: Added a new run method
3. General Cleaning: Made the code a bit more readable for beginners
4. Custom Playback: Added a feature to allow user to choose which file they want to playback when replay is true. (Note user cannot choose the save file name, but they can choose what file to playback by changing the model parameter)
5. Unique_id: Added a unique id to each agent
6. Doc String: Added details of the model class

Discussion Topic:

While it is possible to now replay models, I feel there are still some things that can be improved,
1. It is currently not easy to set custom save name for the checkpoint
2. It is kinda inconvenient to make a new class just to save a file. I think the code could be optimised to make it as simple as something like "pickle.dump". This way it would be easy to save multiple conditional checkpoints in one run. (EDIT: I just noticed it is already under plans in GSOC idea list)


coderbeta1:
pre-commit.ci autofix

examples/caching_and_replay/model.py
```diff
@@ -39,48 +40,64 @@ class Schelling(mesa.Model):
...
         self.height = height
+        self.width = width
         self.density = density
         self.minority_pc = minority_pc
         self.homophily = homophily
+        self.radius = radius
 
         self.schedule = mesa.time.RandomActivation(self)
-        self.grid = mesa.space.SingleGrid(width, height, torus=True)
+        self.grid = mesa.space.SingleGrid(height, width, torus=True)
```
rht, line 74: Shouldn't it be width first?

examples/caching_and_replay/model.py
```diff
@@ -39,48 +40,64 @@ class Schelling(mesa.Model):
...
+            model_reporters={"happy": "happy"},  # Model-level count of happy agents
         )
 
         # Set up agents
         # We use a grid iterator that returns
         # the coordinates of a cell as well as
         # its contents. (coord_iter)
-        for cell in self.grid.coord_iter():
-            x, y = cell[1]
+        for _cont, pos in self.grid.coord_iter():
```
rht, line 85: Unused loop variable should be named just `_`.

examples/caching_and_replay/run.py
```diff
@@ -6,25 +6,30 @@
...
 model_params["replay"] = mesa.visualization.Checkbox("Replay cached run?", False)
+model_params["cache_file_path"] = "my_cache_file_path.cache"
 
 
 def get_cache_file_status(_):
     """
     Display an informational text about caching and the status of the cache file (existing versus not existing)
     """
-    cache_file = Path("./my_cache_file_path.cache")
+    cache_file = Path("./" + model_params["cache_file_path"])
```
rht, line 16: The `./` should be part of `model_params["cache_file_path"]`, given that it is a path instead of a filename.

rht:
Other than the comments, everything else LGTM.

examples/caching_and_replay/model.py
coderbeta1: I tried to follow the benchmark schelling model, and in that height is first. 

But on checking `mesa.space.SingleGrid` width is indeed first. The change needs to be done in the benchmark code also. I have submitted a PR previously for benchmark, I'll make the change in all the schelling models in this repo and the maine mesa repo

PR #104, opened 2024-02-25
Improve Boid Flocking model to benchmark

coderbeta1:
Changes Made:

1. Benchmark: Changed the default values and code to that of [benchmark](https://github.com/coderbeta1/mesa_copy/blob/main/benchmarks/Flocking/flocking.py)

### DIFF
--- a/examples/boid_flockers/boid_flockers/model.py
+++ b/examples/boid_flockers/boid_flockers/model.py
@@ -33,15 +33,15 @@ class Boid(mesa.Agent):
         direction,
         vision,
         separation,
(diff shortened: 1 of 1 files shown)


rht (approved):

rht:
Merged, thank you.

rht:
The one consistent with benchmark has division by N instead of by 2. This is also consistent with the implementation in https://github.com/JuliaDynamics/ABM_Framework_Comparisons, which has the fix https://github.com/JuliaDynamics/ABMFrameworksComparison/pull/73.

PR #106, opened 2024-02-27
Sugarscape Chapter 3- Draft Pull Request

apranav22:
* creating the base code for chapter 3 of sugarscape, attempt to keep as a superset of chapter 2. Ideally, once this example is in a releasable state I would like to remove chapter 2's dedicated example.

TODO:
* Refactor the example code into AgentSets.

Current Issues:
*  Struggling with agent reporters, I'd like to access agents' age data from `agents.py` using the data collector `datacollector` in `models.py` to visualise a bar chart. However, the bar chart does not show up. 
* x----

* On branch sugarscape Changes to be committed:
	new file:   examples/sugarscape_scc/README.md
	new file:   examples/sugarscape_scc/sugarscape_scc/__init__.py
	new file:   examples/sugarscape_scc/sugarscape_scc/resources/ant.png
	new file:   examples/sugarscape_scc/sugarscape_scc/sugar-map.txt
Changes not staged for commit:
	modified:   examples/sugarscape_cg/sugarscape_cg/model.py
Untracked files:
	examples/sugarscape_scc/sugarscape_scc/agents.py
	examples/sugarscape_scc/sugarscape_scc/model.py

* modified:   examples/sugarscape_cg/sugarscape_cg/model.py

* Changes to be committed:
	new file:   examples/sugarscape_scc/sugarscape_scc/model.py
Untracked files:
	examples/sugarscape_scc/sugarscape_scc/agents.py

* new file:   examples/sugarscape_scc/sugarscape_scc/agents.py

* modified:   examples/sugarscape_scc/README.md

* modified:   examples/sugarscape_cg/sugarscape_cg/model.pymodified:   examples/sugarscape_scc/README.md

PR #107, opened 2024-02-27
Follow on commit with minor changes

apranav22:
I don't know the procedure of follow-on commits. Would it be better to keep updating my fork instead in the future?


EwoutH:
Yes, you can just update the branch on your fork and than the PR updates automatically.

Can I close this one that we continue on #106?

apranav22:
Yes, that works. Thanks for the clarification!

Issue #76, opened 2023-12-26
Bug: Some examples weren't updated to 2.0 in their requirements


jackiekazil:
Forrest fire was updated in #75 
This commit shows that there are others where requirements aren't updated, but the code was.
https://github.com/projectmesa/mesa-examples/commit/428c900564d495f725c4cedc0a93b8a7e5463799

Things that need to happen... 

- [ ] Identify which ones had code changed but didn't have requirements updated 
- [ ] Test the code against 2+
- [ ] Update the requirements as appropriate and file bugs for the rest.

PR #109, opened 2024-03-01
fix sugarscape_g1mt running options, requirements and changes to README

FoFFolo:
I fixed the running options on the run.py script according to what the README says, adjusting the option to run "mesa runserver" command.
I also fixed the requirements text file as declared in the issue #76.
I modified the README based on the precedent PRs.
I noticed that there are other problems, especially in the app.py when launching solara, but I don't know if my changes are actually meaningful to merge, so I don't want to do a lot of messy changes.


examples/sugarscape_g1mt/Readme.md
```diff
@@ -69,21 +65,21 @@ To run the model with BatchRunner:
...
+  $ mesa runserver
 ```
 
 Then open your browser to [http://127.0.0.1:8521/](http://127.0.0.1:8521/) and press Reset, then Run.
 
 ## Files
 
 * ``sugarscape_g1mt/trader_agents.py``: Defines the Trader agent class.
-* ``sugarscape_g1mt/resource_agents.py``: Defines the Sugar and Spice agent classes.
+* ``sugarscape_g1mt/resource_agents.py``: Defines the Resource agent class which contains an amount of sugar and spice.
```
rht, line 76: Those all should be single backticks, since they are Markdown syntax.

examples/sugarscape_g1mt/run.py
```diff
@@ -64,42 +64,43 @@ def assess_results(results, single_agent):
 
 args = sys.argv[1:]
 
-
-if args[0] == "runserver":
+try:
```
rht, line 67: Using a check `if len(args) == 0` then `server.launch()` would reduce the number of nesting by 1. According to [PEP 20](https://peps.python.org/pep-0020/), flat is better than nested.

rht:
LGTM

rht:
Merged, thank you @FoFFolo.

PR #114, opened 2024-03-31
el_farol/model: height/width params not necessary

puer-robustus:


rht:
Merged, after modifying the commit message to be imperative, thank you.

PR #113, opened 2024-03-30
fix pkl file name in agents_and_networks example

itsahmedkhalil:
There is no 'outputs' directory. Changed line 70 to: self._path_cache_result = f"{campus}_path_cache_result.pkl".

PR #115, opened 2024-03-31
agents & networks gis model: add missing outputs folder

wang-boyu:
Originally reported in https://github.com/projectmesa/mesa-examples/pull/113 (thanks to @itsahmedkhalil). But I messed up the git history by trying to overwrite original commit with mine (did a hard reset followed by a force push) and the original PR ended closed.

Instead of changing source code, I added an empty `outputs` folder to the example.


EwoutH (approved):
Great solution, thanks!

Issue #136, opened 2024-07-10
Implement Netlogo like Ising Model using GridSpace


vitorfrois:
The Ising Ferromagnet is a simple model with such relevance in Statistical Physics. It can be used to study order, phase-transitions and emergent phenomena. 

_A crystalline material is conceived of as a set of lattice points with spins. The spins (magnetic moments) of the atoms in the magnet can either be up or down. Spins can change as a result of being influenced by neighboring spins and by the ambient temperature. Spins prefer to align with their neighbors. The overall behavior of the system will vary depending on the temperature._

[NetLogo Ising Page](https://ccl.northwestern.edu/netlogo/models/Ising)

vitorfrois:
Ising Model is such an influential model in Opinion Dynamics. Since I'm currently studying the topic, and want to implement this and also get to know better about the new Visualization.

PR #138, opened 2024-07-13
Mesa implementation of Ising Model

vitorfrois:
Ising Model (https://ccl.northwestern.edu/netlogo/models/Ising, https://en.wikipedia.org/wiki/Ising_model) implementation using new Visualization of Ising Model as issued at #136.
image

### DIFF
--- a/examples/ising/README.md
+++ b/examples/ising/README.md
@@ -1,0 +1,2 @@
+# Ising Model
+
--- a/examples/ising/README.md
+++ b/examples/ising/README.md
@@ -1,2 +1,25 @@
 # Ising Model
 
+## Summary
+
+The Ising model (or Lenz–Ising model), named after the physicists Ernst Ising and Wilhelm Lenz, is a mathematical model of ferromagnetism in statistical mechanics. The model consists of discrete variables that represent magnetic dipole moments of atomic "spins" that can be in one of two states (+1 or −1). The spins are arranged in a graph, usually a lattice (where the local structure repeats periodically in all directions), allowing each spin to interact with its neighbors. Neighboring spins that agree have a lower energy than those that disagree; the system tends to the lowest energy but heat disturbs this tendency, thus creating the possibility of different structural phases. The model allows the identification of phase transitions as a simplified model of reality. The two-dimensional square-lattice Ising model is one of the simplest statistical models to show a phase transition.
+
+## How to Run
+
+To run the model interactively, run ``solara run run.py`` in this directory. e.g.
+
+Then open your browser to [http://127.0.0.1:8521/](http://127.0.0.1:8521/) and press ``run``.
+
+## Things to Try
+
+What happens when the temperature slider is very high? (This is called the "paramagnetic" state.) Try this with different **Spin Up Probability** values.
+
+What happens when the temperature slider is set very low? (This is called the "ferromagnetic" state.) Again, try this with different **Spin Up Probability** values.
+
+Between these two very different behaviors is a transition point. On an infinite grid, the transition point can be proved to be $2 / ln (1 + sqrt 2)$, which is about 2.27. On a large enough finite toroidal grid, the transition point is near this number.
+
+## References
+
+(NetLogo Ising)[https://ccl.northwestern.edu/netlogo/models/Ising]
+
+(Introduction to Monte Carlo methods for an Ising Model of a Ferromagnet)[https://arxiv.org/pdf/0803.0217]
--- a/examples/ising/ising/model.py
+++ b/examples/ising/ising/model.py
@@ -5,14 +5,12 @@ from .spin import Spin
 
 
 class IsingModel(mesa.Model):
-    BOLTZMANN_CONSTANT = 1.380649e-23
-
     def __init__(
         self,
         width=50,
         height=50,
-        spin_up_probability: float = 0.5,
-        temperature: float = 2.27,
+        spin_up_probability: float = 0.7,
+        temperature: float = 1,
     ):
         super().__init__()
         self.temperature = temperature
@@ -27,6 +25,7 @@ class IsingModel(mesa.Model):
         self.running = True
 
     def step(self):
+        self._steps += 1
         agents_list = list(self.agents)
         random_spin = self.random.choice(agents_list)
         dE = self.get_energy_change(random_spin)
@@ -34,14 +33,17 @@ class IsingModel(mesa.Model):
             random_spin.state *= -1
         else:
             if self.random.random() < self.boltzmann(dE):
+                print("change spin")
                 random_spin.state *= -1
 
     def get_energy_change(self, spin: Spin):
         neighbors = spin.neighbors()
         sum_over_neighbors = 0
         for neighbor in neighbors:
-            sum_over_neighbors += spin.state * neighbor.state
-        return -1 * sum_over_neighbors
+            sum_over_neighbors += neighbor.state
+        return sum_over_neighbors * 2 * spin.state
 
     def boltzmann(self, dE):
-        return np.exp(-dE / (self.BOLTZMANN_CONSTANT * self.temperature))
+        v = np.exp(-dE / self.temperature)
+        print(v)
+        return v
--- a/examples/ising/ising/model.py
+++ b/examples/ising/ising/model.py
@@ -1,0 +1,47 @@
+import mesa
+import numpy as np
+
+from .spin import Spin
+
+
+class IsingModel(mesa.Model):
+    BOLTZMANN_CONSTANT = 1.380649e-23
+
+    def __init__(
+        self,
+        width=50,
+        height=50,
+        spin_up_probability: float = 0.5,
+        temperature: float = 2.27,
+    ):
+        super().__init__()
+        self.temperature = temperature
+        self.grid = mesa.space.SingleGrid(width, height, torus=True)
+
+        for contents, (x, y) in self.grid.coord_iter():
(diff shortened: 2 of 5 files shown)


examples/ising/ising/model.py
```diff
@@ -0,0 +1,47 @@
...
+                cell.state = cell.UP
+            self.grid.place_agent(cell, (x, y))
+
+        self.running = True
+
+    def step(self):
+        agents_list = list(self.agents)
+        self._steps += 1000
+        for i in range(1000):
+            random_spin = self.random.choice(agents_list)
```
rht, line 31: Having to create a new list of agents from scratch and selecting one is not ideal. I know this is a limitation of `AgentSet`. @EwoutH any idea to do this in a performant way via `AgentSet`? It seems to be a tower of abstractions, that it has slowed down the benchmark for mesa-frames as well: see https://github.com/adamamer20/mesa-frames/issues/25.

examples/ising/ising/portrayal.py
```diff
@@ -0,0 +1,19 @@
...
+    :return: the portrayal dictionary.
+    """
+    if spin is None:
+        raise AssertionError
+    return {
+        "marker": "s",
+        "w": 1,
+        "h": 1,
+        "Filled": "true",
+        "Layer": 0,
```
rht, line 15: This line is not necessary.

examples/ising/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+Then open your browser to [http://127.0.0.1:8521/](http://127.0.0.1:8521/) and press ``run``.
+
+## Things to Try
+
+What happens when the temperature slider is very high? (This is called the "paramagnetic" state.) Try this with different **Spin Up Probability** values.
+
+What happens when the temperature slider is set very low? (This is called the "ferromagnetic" state.) Again, try this with different **Spin Up Probability** values.
+
+Between these two very different behaviors is a transition point. On an infinite grid, the transition point can be proved to be $2 / ln (1 + sqrt 2)$, which is about 2.27. On a large enough finite toroidal grid, the transition point is near this number.
```
rht, line 19: We don't have means of aggregating various param sweep into a single plot of magnetization vs temperature. This is a homework for us developers to allow visualizing batch_run result easily.

rht:
This model is very helpful in demonstrating phase transition. We have Epstein civil violence, but the Ising model is the bread and butter toy model for phase transition in statistical physics. We should aim towards being able to specify this model in code concisely.

examples/ising/ising/model.py
vitorfrois: thanks for the comment. it does not feel right at all haha.

examples/ising/ising/portrayal.py
```diff
@@ -0,0 +1,15 @@
+def portraySpin(spin):
```
rht, line 1: Should be `portray_spin` for the Python convention.

examples/ising/ising/model.py
```diff
@@ -0,0 +1,47 @@
...
+                    random_spin.state *= -1
+
+    def get_energy_change(self, spin: Spin):
+        neighbors = spin.neighbors()
+        sum_over_neighbors = 0
+        for neighbor in neighbors:
+            sum_over_neighbors += neighbor.state
+        return sum_over_neighbors * 2 * spin.state
+
+    def boltzmann(self, dE):
```
rht, line 46: Would be clearer to call this `boltzmann_factor`.

rht:
Other than the 2 comments, this PR LGTM. Once you fix them, I will merge.

examples/ising/ising/model.py
EwoutH: Currently I don’t think there is. @vitorfrois could you open a new issue on the main Mesa repo describing this lack of functionality and linking to this PR as example?

EwoutH: I was going to suggest you can just do:
```Python
self.agents.select(n=1000)
```
But looking at our current implementation ([here](https://github.com/projectmesa/mesa/blob/50b2964e80897777e92935cb1271cce1c3314845/mesa/agent.py#L124)) that isn’t actually random.

What you can do though is:
```Python
self.agents.shuffle().select(n=1000)
```

examples/ising/README.md
EwoutH: Agreed. Some related tickets:

- https://github.com/projectmesa/mesa/discussions/1970
- https://github.com/projectmesa/mesa/issues/2055
- The famous https://github.com/projectmesa/mesa/issues/348

examples/ising/run.py
EwoutH, line 1: For consistency is prefer this named `app.py`.

It might not be the best name, but it’s what we use now.

examples/ising/run.py
```diff
@@ -0,0 +1,28 @@
+import solara
+from ising.model import IsingModel
+from ising.portrayal import portray_spin
+from mesa.visualization import JupyterViz
```
EwoutH, line 4: You can now import SolaraViz
```suggestion
from mesa.visualization import SolaraViz
```

examples/ising/run.py
```diff
@@ -0,0 +1,28 @@
...
+        min=0,
+        max=1,
+        step=0.05,
+    ),
+}
+
+
+@solara.component
+def Page():
+    JupyterViz(
```
EwoutH, line 23:
```suggestion
    SolaraViz(
```

examples/ising/ising/spin.py
```diff
@@ -0,0 +1,19 @@
...
+    def __init__(self, pos, model, init_state):
+        """
+        Create a cell, in the given state, at the given x, y position.
+        """
+        super().__init__(pos, model)
+        self.x, self.y = pos
+        self.state = init_state
+
+    def neighbors(self):
+        return self.model.grid.iter_neighbors((self.x, self.y), True)
```
EwoutH, line 19: Make explicit for readability:
```suggestion
        return self.model.grid.iter_neighbors((self.x, self.y), moore=True)
```

examples/ising/ising/spin.py
```diff
@@ -0,0 +1,19 @@
...
+    """Represents a single ALIVE or DEAD cell in the simulation."""
+
+    UP = 1
+    DOWN = -1
+
+    def __init__(self, pos, model, init_state):
+        """
+        Create a cell, in the given state, at the given x, y position.
+        """
+        super().__init__(pos, model)
```
EwoutH, line 14: Doesn’t this map to `unique_id, model`?

If intended, please add a comment, if not, update

examples/ising/ising/portrayal.py
```diff
@@ -0,0 +1,15 @@
...
+    """
+    This function is registered with the visualization server to be called
+    each tick to indicate how to draw the cell in its current state.
+    :param cell:  the cell in the simulation
+    :return: the portrayal dictionary.
+    """
+    if spin is None:
+        raise AssertionError
+    return {
+        "marker": "s",
```
EwoutH, line 11: Already using the new marker function, awesome!

I think we named it shape though: https://github.com/projectmesa/mesa/pull/2214

```suggestion
        "shape": "s",
```
Or if this intended?

EwoutH: CC @rmhopkins4

examples/ising/ising/model.py
```diff
@@ -0,0 +1,47 @@
...
+
+        self.running = True
+
+    def step(self):
+        agents_list = list(self.agents)
+        self._steps += 1000
+        for i in range(1000):
+            random_spin = self.random.choice(agents_list)
+            dE = self.get_energy_change(random_spin)
+            if dE < 0:
```
EwoutH, line 33: Maybe you can do here
```suggestion
            if dE < 0 or self.random.random() < self.boltzmann_factor(dE):
```
If I understand Python correctly, the later part won’t be evaluated if the former is `False` already, making it just as fast, and more readable.

EwoutH (commented):
It looks like a really clean and neat implementation! I already like it a lot, especially that it demonstrates some more recent Mesa functionality and comes with visualization.

I left some comments, but already like it a lot!

examples/ising/ising/model.py
```diff
@@ -0,0 +1,47 @@
...
+            cell = Spin((x, y), self, Spin.DOWN)
+            if self.random.random() < spin_up_probability:
+                cell.state = cell.UP
+            self.grid.place_agent(cell, (x, y))
+
+        self.running = True
+
+    def step(self):
+        agents_list = list(self.agents)
+        self._steps += 1000
```
quaquel, line 29: If I understand this correctly, you are using _step to count the number of agent activations (you use 1000, as in the for loop). Why use `_step` for this?

EwoutH:
@vitorfrois I still think this is awesome example. Can I help you finish the last parts of this PR?

examples/ising/run.py
vitorfrois: I dont know if I understood how to update the visualization correctly. are we using the Streamlit platform for default now?

examples/ising/ising/spin.py
vitorfrois: pos is a tuple `(x,y)`, so I think it makes more sense to name it like this instead of unique_id. I updated the name to `position` to be more informative

EwoutH:
We made some breaking changes in the latest Mesa 3.0 pre-release, but our Migration guide is now fully up to date, which might help: https://mesa.readthedocs.io/en/latest/migration_guide.html

Issue #139, opened 2024-07-16
A website for example models?


wang-boyu:
### Discussed in https://github.com/projectmesa/mesa-examples/discussions/61

<div type='discussions-op-text'>

<sup>Originally posted by **wang-boyu** September 21, 2023</sup>
I came across with these websites in one of the classes I take this semester: [isee Exchange™](https://exchange.iseesystems.com) and [anylogic cloud](https://cloud.anylogic.com/models). They include models from both the official team as well as the community. Models can be run from within the websites, and there are statistics showing number of runs etc. NetLogo also has its [NetLogo Web](http://www.netlogoweb.org/launch#http://www.netlogoweb.org/assets/modelslib/Sample%20Models/) (but without the statistics). I really like the run statistics as they provide a kind of feedback from users on which models are more popular.

This makes me wonder whether we can have something similar for Mesa example models. One advantage we have over Stella/AnyLogic/NetLogo is that we can rely on Google Colab to run models using Jupyter Notebooks. This would also be in line with what we have for Mesa & Mesa-Geo, where there are readthedocs sites for users and GitHub repos for developers. Similarly there could be a website for example model explorers and this GitHub repo containing all source code of the example models.</div>

EwoutH:
I think pycafe is suggested as one of the alternatives.

- https://github.com/projectmesa/mesa/discussions/2170

wang-boyu:
I tried to use Solara to build something like https://solara.dev/documentation/examples for Mesa examples. With some kind help from @iisakkirotko in [here](https://github.com/widgetti/solara/discussions/821) and Solara's souce code for the page, I was able to get this:

<img width="1056" alt="Screenshot 2024-10-17 at 5 46 45 PM" src="https://github.com/user-attachments/assets/2054f467-8dc3-4afe-948d-086c50f50548">

Source code is at my fork: https://github.com/wang-boyu/mesa-examples/tree/mesa-website. The command used is simply `solara run mesa_models.website.pages` to get it running locally.

My takeaways so far:

1. This needs to be deployed in some (paid?) servers.
2. Documentation may be moved from ReadtheDocs to here if we have such server, much like Solara documentation: https://solara.dev/documentation
3. It would also be nice to have some kind of user management system, so that users can create accounts and we can track popularity of models (number of plays, number of likes, etc), like what's in AnyLogic: https://cloud.anylogic.com/models
4. For each model we can probably add links to open it in Google Colab or pycafe

[...]

PR #140, opened 2024-07-16
Categorize example models into groups

wang-boyu:
This PR groups Mesa example models into different groups based on the types of space (e.g., grid, continuous, network, etc) in the README file.


rht (approved):

EwoutH:
Great improvement!

Another thing I was thinking about was making some kind of table / list which models are using which grids, scheduler, etc.

wang-boyu:
Thanks! If we have a model website (mentioned in https://github.com/projectmesa/mesa-examples/issues/139) maybe we can have some kind of options to filter models by types.

PR #141, opened 2024-07-21
merge README files into top level README

wang-boyu:
- Move two README files from sub folders into top level README file
- Create a table of contents for all example categories


EwoutH (approved):
Good idea, this makes it directly visible.

PR #155, opened 2024-08-11
Use SolaraViz instead of JupyterViz in examples

EwoutH:
This updates all the examples to use the renamed and stabilized SolaraViz.

For reference:
- https://github.com/projectmesa/mesa/pull/2090
- https://github.com/projectmesa/mesa/pull/2188

### DIFF
--- a/examples/schelling_experimental/app.py
+++ b/examples/schelling_experimental/app.py
@@ -1,4 +1,4 @@
-from mesa.visualization.jupyter_viz import JupyterViz, Slider, make_text
+from mesa.visualization.solara_viz.py import Slider, SolaraViz, make_text
 from model import Schelling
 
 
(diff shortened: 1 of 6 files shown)


rht (approved):

EwoutH:
Thanks for merging!

examples/schelling_experimental/app.py
```diff
@@ -1,4 +1,4 @@
-from mesa.visualization.jupyter_viz import JupyterViz, Slider, make_text
+from mesa.visualization.solara_viz.py import Slider, SolaraViz, make_text
```
rht, line 1: This line doesn't work.

EwoutH: Do you think because of the re-order? Ruff mandated that.

rht: Yes, removing .py fixed it.

EwoutH: Fixed in 274a1bcdce13d86f992eed085093f8954400ad98

PR #158, opened 2024-08-13
Epstein civil violence: Switch to experimental datacollector

rht:


examples/epstein_civil_violence/epstein_civil_violence/model.py
```diff
@@ -102,16 +86,40 @@ def __init__(
...
-        self.running = True
-        self.datacollector.collect(self)
+        model_reporters = {
+            "Quiescent": lambda m: self.count_type_citizens(m, "Quiescent"),
+            "Active": lambda m: self.count_type_citizens(m, "Active"),
+            "Jailed": self.count_jailed,
+            "Cops": self.count_cops,
+        }
+        agent_reporters = {
+            "x": lambda agents: [pos[0] for pos in agents.get("pos")],
```
rht, line 96: This API is awkward. It is better to use instead `lambda a: a.pos[0]`. But this means `AgentSet` needs to have the feature of `.apply()` along the line of DataFrame's apply.

This also means that the new `DataCollector` heavily relies on `AgentSet` in order to remain simple. Simple is good.

EwoutH: I think @quaquel was also thinking in this direction: https://github.com/projectmesa/mesa/discussions/1944#discussioncomment-8490396

quaquel: I am back in office and do have some time. However, I lot has been happening wrt to MESA so @EwoutH if you have 30 minutes to bring me up to speed, I am happy to take a look at the apply idea.

EwoutH: Good to hear! Let me get back to you what works.

EwoutH: @quaquel would tomorrow early afternoon work?

quaquel: yes, just email me a time and place (online or my office)

Issue #111, opened 2024-03-19
Use new AgentSet functionality in examples


EwoutH:
_Open for contributions!_

I would like some example to use the now stabilized AgentSet. This means sometimes self.schedule.agents can be replaced, and sometimes the scheduler could be removed all together. See:

- Original PR: https://github.com/projectmesa/mesa/pull/1916
- Current docs: https://mesa.readthedocs.io/en/stable/mesa.html#mesa.agent.AgentSet
- Original discussion: https://github.com/projectmesa/mesa/discussions/1912

The approach for this would generally be:

- Read up on AgentSet functionality
- Go through the Mesa examples and seek oppertunities where the AgentSet could simplify or improve things.
- Open a Discussion on mesa-examples proposing your changes
- Implement them and open a PR here

The nuance here being that applying the AgentSet is not a goal to be implemented as much as possible, but a means that could make some model (parts) more elegant, efficient, readable and/or performant.

EwoutH:
In https://github.com/projectmesa/mesa-examples/pull/153 I removed some schedulers and replaced them with AgentSet functionality. It could be done this way for more models!

EwoutH:
With Mesa [v3.0 alpha 2](https://github.com/projectmesa/mesa/releases/tag/v3.0.0a2) (`3.0.0a2`) merged, we can now also start using `AgentSet.do(method=Callable)` in our examples! See https://github.com/projectmesa/mesa/pull/2219.

Rishi2674:
Hey @EwoutH , I am reading some resources you gave above for this issue. Can I work on this one?

sakshirakate11:
Hi @EwoutH,

I am a beginner contributor and very interested in working on this issue.
I am currently exploring Mesa documentation and AgentSet functionality.

I would like to work on this issue and propose improvements to Mesa examples using AgentSet.

Please assign this issue to me if possible.

Thank you!

ImAnouar:
Hi @EwoutH , I'd like to work on a small AgentSet improvement for the Deffuant–Weisbuch example under this issue. I’m currently using Mesa in my research on opinion dynamics.

Please let me know if I can proceed.

Thanks!

PR #161, opened 2024-08-17
Replace RandomActivation and SimultaneousActivation schedulers with AgentSet functionality

EwoutH:
This PR replace all the `RandomActivation` and `SimultaneousActivation` schedulers used in the example models with AgentSet functionality. This simplifies models and allows for easier and more flexible modification of these models. It algins the examples with the current best practices in Mesa.

### RandomActivation (b7f67c7)
The RandomActivation schedulers were removed, and in the Model.step() function the schedule step was replaced with an AgentSet operation, like:

```diff
-        self.schedule.step()
+        self.agents.shuffle().do("step")
```

The following RandomActivation schedulers were removed:
- bank_reserves
- boid_flockers
- boltzmann_wealth_model
- boltzmann_wealth_model_experimental
- boltzmann_wealth_model_network
- caching_and_replay
- charts
- el_farol
- epstein_civil_violence
- forest_fire
- hotelling_law
- schelling
- shape_example
- virus_on_network

### SimultaneousActivation (7e396fe)
To remove the SimultaneousActivation schedulers, the Agent methods were renamed from `step()` and `advance()` to more descriptive functions. The model step was updated accordingly:

```diff
-        self.schedule.step()
+        self.agents.do("determine_state")
+        self.agents.do("assume_state")
```

The following SimultaneousActivation schedulers were removed:
- hex_snowflake
- color_patches
- conways_game_of_life

**Relevant tickets**
- https://github.com/projectmesa/mesa-examples/issues/111
- https://github.com/projectmesa/mesa/discussions/1912
- https://github.com/projectmesa/mesa/pull/1916

### DIFF
--- a/examples/bank_reserves/bank_reserves/agents.py
+++ b/examples/bank_reserves/bank_reserves/agents.py
@@ -14,6 +14,12 @@ from .random_walk import RandomWalker
 
 
 class Bank:
+    """Note that the Bank class is not a Mesa Agent, but just a regular Python
+    class. This is because there is only one bank in this model, and it does not
+    use any Mesa-specific features like the scheduler or the grid, and doesn't
+    have a step method. It is just used to keep track of the bank's reserves and
+    the amount it can loan out, for Person agents to interact with."""
+
     def __init__(self, model, reserve_percent=50):
         self.model = model
         # for tracking total value of loans outstanding
--- a/examples/bank_reserves/bank_reserves/agents.py
+++ b/examples/bank_reserves/bank_reserves/agents.py
@@ -10,15 +10,12 @@ Author of NetLogo code:
     Northwestern University, Evanston, IL.
 """
 
-import mesa
-
 from .random_walk import RandomWalker
 
 
-class Bank(mesa.Agent):
-    def __init__(self, unique_id, model, reserve_percent=50):
-        # initialize the parent class with required parameters
-        super().__init__(unique_id, model)
+class Bank:
+    def __init__(self, model, reserve_percent=50):
+        self.model = model
         # for tracking total value of loans outstanding
         self.bank_loans = 0
         """percent of deposits the bank must keep in reserves - this is set via
--- a/examples/bank_reserves/bank_reserves/agents.py
+++ b/examples/bank_reserves/bank_reserves/agents.py
@@ -173,7 +173,6 @@ class Person(RandomWalker):
         # increase the bank's outstanding loans
         self.bank.bank_loans += amount
 
-    # step is called for each agent in model.BankReservesModel.schedule.step()
     def step(self):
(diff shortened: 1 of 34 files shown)


EwoutH:
And believe it or not, the final error is the DataCollector that tries to gather attributes from agents that don't have said attributes.

EwoutH:
I updated the PR description and title and cleaned up the commit history as far as feasible.

Looking forward to reviews!

examples/bank_reserves/bank_reserves/agents.py
```diff
@@ -10,15 +10,12 @@
...
 
-import mesa
-
 from .random_walk import RandomWalker
 
 
-class Bank(mesa.Agent):
-    def __init__(self, unique_id, model, reserve_percent=50):
-        # initialize the parent class with required parameters
-        super().__init__(unique_id, model)
```
rht, line 21: Why is the `super().__init__` removed? There could be other attributes initialized other than the UID and model.

rht: `mesa.Agent` `__init__` has
```python
self.model.agents_[type(self)][self] = None
```
without the `super().__init__`, the agent is not added to the `AgentSet`.

examples/charts/charts/agents.py
```diff
@@ -10,15 +10,12 @@
...
 
-import mesa
-
 from .random_walk import RandomWalker
 
 
-class Bank(mesa.Agent):
-    def __init__(self, unique_id, model, reserve_percent=50):
-        # initialize the parent class with required parameters
-        super().__init__(unique_id, model)
```
rht, line 21: This one as well.

rht:
Other than the 2 comments, everything else LGTM. Feel free to rebase and merge after they had been addressed.

examples/bank_reserves/bank_reserves/agents.py
EwoutH: I explained it briefly in the commit message:

> bank_reserves/charts: Don't make Bank an Agent
> It doesn't have a step function and there is only one, so it shouldn't be an Agent, just an Python object.

The idea is that the bank doesn't need to be an Agent, it can just be an Python object. No agent functionality is used, and it doesn't have a step function, so it can just be an object on which other agents interact on.

rht: Probably should be added in the docstring of the `Bank` class.

EwoutH: That's a good point, I added an extended docstring note!

rht (approved):

EwoutH:
With rht’s approval I’m going to merge this.

@jackiekazil and @tpike3 I would still love your take, if you have any feedback let me know.

PR #162, opened 2024-08-18
schelling_experimental: Remove temporary _advance_time() method

EwoutH:
Remove the remove temporary _advance_time() method from schelling_experimental. This isn't needed anymore, because in Mesa 3.0 time will be automatically increased.

See https://github.com/projectmesa/mesa/pull/2223.


rht:
Reminder to always delete the branch after merge or enable branch auto delete in the settings (I don't have the permission to do this).

Issue #168, opened 2024-08-21
virus_on_network new visualization doesn't update while running model.


DrEntropy:
Virus_on_network doesn't update the visualization (using the new SolaraViz) using the current main branch of mesa.  This is due to SolaraViz watching model._steps which is not updated in this model (it doesn't use a built in schedular).

EwoutH:
You’re correct. I expected https://github.com/projectmesa/mesa/pull/2223 to go through together with that, but it’s taking some longer time than expected (see https://github.com/projectmesa/mesa/discussions/2227).

EwoutH:
I've reverted #161 in #170, could you confirm the visualisation now works again?

Thanks for the feedback by the way, I hope we will resolve the conceptual discussions around time at some point and be able to revert-revert #161 maybe even.

DrEntropy:
Yes I have confirmed that reverting to using schedulers fixes the issue, thanks!

PR #169, opened 2024-08-21
Fix [issue #168] virus_on_network not updating

DrEntropy:
As stated in [#168](https://github.com/projectmesa/mesa-examples/issues/168)   virus_on_network visualization does not update as the model advances. The issue is that SolaraViz watches `model._step` which doesn't advance in this model because the model doesn't use a scheduler.   This is fixed by calling `self._advance_time()` in `model:step()`.    This is a one line change.

### DIFF
--- a/examples/virus_on_network/virus_on_network/model.py
+++ b/examples/virus_on_network/virus_on_network/model.py
@@ -97,6 +97,8 @@ class VirusOnNetwork(mesa.Model):
 
     def step(self):
         self.agents.shuffle().do("step")
+        # advance model time
+        self._advance_time()
         # collect data
         self.datacollector.collect(self)
 


EwoutH:
Thanks for the PR. I’m holding of with this PR until we have a bit more clarity on https://github.com/projectmesa/mesa/pull/2223 / https://github.com/projectmesa/mesa/discussions/2227.

If it isn’t resolved next week, we’re either need to revert https://github.com/projectmesa/mesa-examples/pull/161 or do this change for all models.

I hope it doesn’t come to that. Feel free to join any of the discussions.

EwoutH:
Thanks for the PR, we've reverted #161 (done in #170), so we're back to the schedulers right now, until we have our conceptual view of time fully sorted out.

Thanks anyways!

Issue #172, opened 2024-08-25
Fix all GIS models


EwoutH:
We just started testing our GIS models in CI (see https://github.com/projectmesa/mesa-examples/pull/171), and currently unfortunately all 7 models that are tested present one or multiple errors

The errors are visible in any recent CI run of the [Test GIS models](https://github.com/projectmesa/mesa-examples/actions/workflows/test_gis_examples.yml) workflow. They can also be reproduced and tested locally by running the `test_gis_examples.py` script with pytest.

![image](https://github.com/user-attachments/assets/79b13a25-c647-4a5e-a511-0af9e90fd707)

#### Current errors
- [x] **GeoSchelling**
  - [x] Fix missing file: `data/nuts_rg_60M_2013_lvl_2.geojson`
  - [x] Fix missing Model initialization
  - [x] UserWarning: The weights matrix is not fully connected:
- [x] **GeoSchellingPoints**
  - [x] Fix missing file: `data/nuts_rg_60M_2013_lvl_2.geojson`
- [x] **Rainfall**
  - [x] Fix missing file: `/vsigzip/data/elevation.asc.gz`
  - [x] Fix missing Model initialization
  - [x] RasterLayer doesn't have access to Model ([mesa-geo#236](https://github.com/projectmesa/mesa-geo/issues/236))
- [x] **Population**
  - [x] Fix missing file: `/vsizip/data/clip.zip`
  - [x] Fix missing Model initialization
  - [x] RasterLayer doesn't have access to Model ([mesa-geo#236](https://github.com/projectmesa/mesa-geo/issues/236))
- [x] **AgentsAndNetworks**
  - [x] Provide the following required arguments (6312ebe75af363ce711ce8c07a70830c076e4fd5):
    - [x] `campus`
    - [x] `data_crs`
    - [x] `buildings_file`
    - [x] `walkway_file`
    - [x] `lakes_file`
    - [x] `rivers_file`
    - [x] `driveway_file`
    - [x] `num_commuters`
  - [x] unhashable type error (f24626e4dac359fd3d3c23e3db8e154497140cd8)
  - [x] set output_dir (679b677c42767b78244e89cf34cc1eb8e1c2622a)
- [x] **GeoSir**
  - [x] Fix missing file: `data/TorontoNeighbourhoods.geojson`
  - [x] Fix missing Model initialization
- [x] **UrbanGrowth**
  - [x] Provide missing argument for `RasterLayer.__init__()`: `model`

[...]

PR #176, opened 2024-08-25
Fix 3 data path imports in gis models

EwoutH:
Fix the data path import errors in the geo_schelling, geo_schelling_points and geo_sir models.

The problem was that many models are called from different place, and might be imported or not. The `os.path.dirname(os.path.abspath(__file__))` part gives the absolute path of the current file and then the directory it's in. From there we can robustly move to the location where the data is located, no matter if we're running locally or not.

Part of https://github.com/projectmesa/mesa-examples/issues/172.

### DIFF
--- a/gis/geo_schelling/model.py
+++ b/gis/geo_schelling/model.py
@@ -1,7 +1,10 @@
+import os
 import random
 
 import mesa
 import mesa_geo as mg
+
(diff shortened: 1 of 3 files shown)


rht:
Yes, fine with me. Though you could use `Path` object instead, with `/` as joiner.

PR #177, opened 2024-08-25
Fix data paths and model passing in GIS examples

EwoutH:
This PR addresses several of the remaining issues across the GIS examples:

- Corrects data file paths in the population, rainfall, and urban growth models to use relative paths from the script directory.
- Fixes the urban growth model by passing the model object to City and UrbanCell classes, as required by mesa-geo 0.8.0.
- Updates the population model to pass the model object when creating agents, ensuring they can access Model variables and properties.
- Corrects the use of `/vsigzip/` paths for GDAL compatibility.

The `/vsigzip/` situation is super weird. It isn't part of the file path, but it's a "marker" a GDAL convention. See https://github.com/projectmesa/mesa-geo/issues/235 for more details. 

Part of #172.

### DIFF
--- a/gis/population/population/model.py
+++ b/gis/population/population/model.py
@@ -1,3 +1,4 @@
+import os
 import math
 import random
 import uuid
@@ -8,6 +9,8 @@ import numpy as np
 from shapely.geometry import Point
 
 from .space import UgandaArea
+
+script_directory = os.path.dirname(os.path.abspath(__file__))
 
 
 class Person(mg.GeoAgent):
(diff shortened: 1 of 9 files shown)


EwoutH:
This PR brings the score for the 7 gis examples from 2 fully passed and 1 passed with warnings, to 3 fully passed and 3 passes with warnings.

EwoutH:
Another two models pass. 5 pass, one warning, one fail.

Those will be separate PRs. This one is ready for review.

gis/population/population/model.py
```diff
@@ -9,6 +10,8 @@
 
 from .space import UgandaArea
 
+script_directory = os.path.dirname(os.path.abspath(__file__))
```
rht, line 13: It's better to use `pathlib.Path` here so that later you can do `script_directory / stuff`.

rht: `script_directory = Path(__file__).resolve().parent`.

EwoutH: Thanks, I looked into it and it indeed is way cleaner!

I used `Path(__file__).resolve().parent` to get the directory of the script.

Implemented in `bcc44b7`.

EwoutH:
I noticed one other issue:
- https://github.com/projectmesa/mesa-geo/issues/236

But that isn't fixable from the examples alone, so something for @wang-boyu or @tpike3 to look into.

I updated the tracking issues https://github.com/projectmesa/mesa-examples/issues/172. Going to merge this PR.

PR #181, opened 2024-08-28
Ensure grasspatches info is only collected if gras is true

quaquel:
the datacollector allways includes grasspatches even if grass is False. This is breaking (the proposed changes to agent storage in Model)[https://github.com/projectmesa/mesa/pull/2251#issue-2488519860].

The underlying problem is that `Model.get_agents_of_type` now raises a `KeyError` if type does not exist. I think this is desirable behavior becuase errrors should not be passed over in silence. But this requires fixing the wolf sheep example as done here.


EwoutH (approved):
Good catch. I think this actually can be a single dict comprehension but let's go for readability here.

PR #184, opened 2024-09-01
gis: Use gzip.open opener in RasterLayer.from_file

EwoutH:
Use pass the new `rio_opener` argument with `gzip.open` in `RasterLayer.from_file()`. This prevents having to do the weird `/vsigzip/` stuff.

Note that urban_growth uses the rasterio opener directly, while the other two examples pass it via the mesa_geo.raster_layers.RasterLayer.open_file() method.

- Follow up on https://github.com/projectmesa/mesa-examples/pull/177
- Enabled by https://github.com/projectmesa/mesa-geo/pull/237
- Part of https://github.com/projectmesa/mesa-geo/issues/235
- Part of https://github.com/projectmesa/mesa-examples/issues/172

### DIFF
--- a/gis/population/population/space.py
+++ b/gis/population/population/space.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import gzip
 import uuid
 
 import geopandas as gpd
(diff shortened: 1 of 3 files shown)


EwoutH:
Just checked the CI, on the main branch this works perfectly!

PR #187, opened 2024-09-02
fix tests for agents-networks gis example

wang-boyu:
Addresses https://github.com/projectmesa/mesa-examples/issues/172#issuecomment-2324471001 as part of https://github.com/projectmesa/mesa-examples/issues/172.

### DIFF
--- a/gis/agents_and_networks/src/model/model.py
+++ b/gis/agents_and_networks/src/model/model.py
@@ -69,6 +69,7 @@ class AgentsAndNetworks(mesa.Model):
         lakes_file=script_directory / "../../data/ub/hydrop.zip",
         rivers_file=script_directory / "../../data/ub/hydrol.zip",
         driveway_file=script_directory / "../../data/data/ub/UB_Rds.zip",
+        output_dir=script_directory / "../../outputs",
         num_commuters=50,
         commuter_min_friends=5,
         commuter_max_friends=10,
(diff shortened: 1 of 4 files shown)


EwoutH (approved):
Thanks! It seems like it’s uncovering a new issue with the AgentsAndNetworks model, but this one is fixed!

gis/agents_and_networks/src/model/model.py
```diff
@@ -59,14 +59,14 @@ class AgentsAndNetworks(mesa.Model):
...
-        data_crs: str,
-        buildings_file: str,
-        walkway_file: str,
-        lakes_file: str,
-        rivers_file: str,
-        driveway_file: str,
-        num_commuters,
+        campus="ub",
+        data_crs="epsg:4326",
+        buildings_file="data/ub/UB_bld.zip",
```
EwoutH, line 64: Actually, the new error might be related:

> E   pyogrio.errors.DataSourceError: '/vsizip/data/ub/UB_bld.zip' does not exist in the file system, and is not recognized as a supported dataset name.

EwoutH:
It seems super().init needs to be called in the AgentsAndNetworks’s Agent

wang-boyu:
For the Building agent: https://github.com/projectmesa/mesa-examples/blob/89f9379d40da872dbef68a4f4b608697845cf00d/gis/agents_and_networks/src/agent/building.py#L23

For model: https://github.com/projectmesa/mesa-examples/blob/89f9379d40da872dbef68a4f4b608697845cf00d/gis/agents_and_networks/src/model/model.py#L81

Am I missing something?

EwoutH:
Something weird is happening here. 

> This error occurs because Python’s dictionary keys must be hashable, and instances of a class are not hashable by default unless you define a __hash__ method in the class.
> 
> In your code, you are trying to use an instance of Building as a key in a dictionary.

But let’s solve that in the next PR. This one already solves two issues!

EwoutH (approved):

wang-boyu:
Probably because a `__eq__()` method is defined for Building but not a `__hash__()` method.

wang-boyu:
Great! Another error to fix : )

wang-boyu:
Should be working now

EwoutH:
Awesome! Thanks for your persistence!

PR #189, opened 2024-09-03
gis: fix user warning when testing geo schelling example

wang-boyu:
As part of https://github.com/projectmesa/mesa-examples/issues/172.

### DIFF
--- a/gis/geo_schelling/model.py
+++ b/gis/geo_schelling/model.py
@@ -1,10 +1,27 @@
 import random
 from pathlib import Path
 
+import geopandas as gpd
+import libpysal
 import mesa
 import mesa_geo as mg
 
 script_directory = Path(__file__).resolve().parent
+
+
+def get_largest_connected_components(gdf):
+    """Get the largest connected component of a GeoDataFrame."""
+    # create spatial weights matrix
+    W = libpysal.weights.Queen.from_dataframe(
+        gdf, use_index=True, silence_warnings=True
+    )
+    # get component labels
+    gdf["component"] = W.component_labels
+    # get the largest component
+    largest_component = gdf["component"].value_counts().idxmax()
+    # subset the GeoDataFrame
+    gdf = gdf[gdf["component"] == largest_component]
+    return gdf
 
 
 class SchellingAgent(mg.GeoAgent):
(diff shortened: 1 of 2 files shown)


wang-boyu:
pre-commit.ci autofix

EwoutH:
Thanks! Could you explain briefly what the warning was and this fixes it?

wang-boyu:
Sure! Sorry that I didn't provide more info.

Basically the raw data contains a set of polygons, and we want to create a graph out of it, where a node represents a polygon, and a link between two polygons means that they are touch neighbors (or in other words, adjacent polygons).

But the polygons in our data can be divided into several groups, and there's no link in-between these groups. I simply keep the largest group for the model.

In `server.py` we hardcoded the `view` and `zoom` parameters to be `[52, 12]` and `4` to zoom into these regions and ignored other regions. Now that we removed other regions, these parameters aren't needed any longer.

Alternatively, we could set `silence_warning` to `True` (links: [here](https://pysal.org/libpysal/generated/libpysal.weights.W.html) and [here](https://pysal.org/libpysal/generated/libpysal.weights.Queen.html#libpysal.weights.Queen.from_iterable)) in Mesa-Geo: https://github.com/projectmesa/mesa-geo/blob/5fb2dde5e63bbfea10775af988848dbfde4ce8a2/mesa_geo/geospace.py#L297. This will silence all warnings for everyone, but I think keeping it may be useful. Hence I fixed it only for the test case of this example.

EwoutH:
Awesome! Could you rebase/update this PR? If correct, this is the one that will turn it green!

If you find the current warnings settings too strict for Mesa-geo, we could also remove the `-Werror` or allow some more warnings to pass.

EwoutH (approved):

wang-boyu:
pre-commit.ci autofix

wang-boyu:
Rebased!

We can probably keep the current warning settings for now and decide later: )

wang-boyu:
Need to rush to a meeting and will check this later

EwoutH:
Green!!!

Awesome.

wang-boyu:
Fixed pre-commit error and squashed all commits into one.

EwoutH:
Go ahead, I approved already

wang-boyu:
Thanks @EwoutH

PR #190, opened 2024-09-03
Replace `get_agents_of_type` method with `agents_by_type` property

EwoutH:
This PR updates the examples now that this PR is merged:
- https://github.com/projectmesa/mesa/pull/2267

The Model method `get_agents_of_type()` is replaced by the `agents_by_type` property, which directly returns the dict.

Instead of using:
```Python
model.get_agents_of_type(Wolf)
```
You should now use:
```Python
model.agents_by_type[Wolf]
```


EwoutH:
It's expected that build-pre fails, since the latest pre-release doesn't have that method yet. build-main passes.

EwoutH:
I'm going to merge this one to stop our main CI from failing.

If you have any comments please still let me know!

PR #193, opened 2024-09-04
update examples to use new solara_viz

Corvince:
WIP to update the models to use the new solaraviz API.

This helps me to write a migration guide for things I notice. So far it goes smoothly, but is a bit tedious

### DIFF
--- a/examples/aco_tsp/app.py
+++ b/examples/aco_tsp/app.py
@@ -4,9 +4,10 @@ Configure visualization elements and instantiate a server
 
 import networkx as nx
 import solara
+from mesa.visualization.utils import update_counter
 from aco_tsp.model import AcoTspModel, TSPGraph
 from matplotlib.figure import Figure
-from mesa.visualization import SolaraViz
+from mesa.visualization import SolaraViz, make_plot_measure
 
 
 def circle_portrayal_example(agent):
@@ -35,6 +36,8 @@ model_params = {
     },
 }
 
+model = AcoTspModel()
+
 
 def make_graph(model):
     fig = Figure()
@@ -55,7 +58,7 @@ def make_graph(model):
         edge_color="gray",
     )
 
-    solara.FigureMatplotlib(fig)
+    return solara.FigureMatplotlib(fig)
 
 
 def ant_level_distances(model):
@@ -67,10 +70,8 @@ def ant_level_distances(model):
 
 
 page = SolaraViz(
-    AcoTspModel,
-    model_params,
-    space_drawer=None,
-    measures=["best_distance_iter", "best_distance", make_graph],
-    agent_portrayal=circle_portrayal_example,
+    model,
+    components=[make_plot_measure(["best_distance_iter", "best_distance"]), make_graph],
+    model_params=model_params,
     play_interval=1,
 )
--- a/examples/aco_tsp/app.py
+++ b/examples/aco_tsp/app.py
@@ -4,7 +4,6 @@ Configure visualization elements and instantiate a server
 
 import networkx as nx
 import solara
-from mesa.visualization.utils import update_counter
 from aco_tsp.model import AcoTspModel, TSPGraph
 from matplotlib.figure import Figure
 from mesa.visualization import SolaraViz, make_plot_measure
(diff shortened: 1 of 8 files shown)


examples/aco_tsp/app.py
```diff
@@ -35,6 +36,8 @@ def circle_portrayal_example(agent):
     },
 }
 
+model = AcoTspModel()
```
EwoutH, line 39: Normally I try to avoid using `model` as variable name for Model instances, since it can lead to both namespace issues (when importing from `model`) and just general confusion.

I often use `model1` or `<name>_model`.

examples/boid_flockers/app.py
```diff
@@ -15,11 +15,12 @@ def boid_draw(agent):
     "separation": 2,
 }
 
+model = BoidFlockers(100, 100, 100, 5, 10, 2)
```
EwoutH, line 18: Adding this here should allow you to reduce the `model_params` dict above, right?

examples/aco_tsp/app.py
Corvince: Interesting, never thought of that really. We should have a best practice for this

examples/boid_flockers/app.py
Corvince: You're right! 

Haven't thought about this. Currently it actually requires all parameters, but we should reduce that to only the parameters that can be changed. Should actually simplify the logic

examples/aco_tsp/app.py
EwoutH: I think I had an multiple hours long debugging session a few years ago, that concluded this. Since then all my instances get an `1` attached unconsciously.

Just now remembered why I did that

examples/boid_flockers/app.py
Corvince: Unfortunately its not so easy. On a parameter change we have to re-initialize the model. But for that we need to know the parameter values of the model initialization. But they are not retained after initialization. Lets think about this a bit more

EwoutH: Sounds like we need 
- https://github.com/projectmesa/mesa/issues/2268

Corvince:
Finally found some time to update the remaining models. There are still a lot of examples that use the old (tornado based) visualizations though. But they are out of scope of this pr

examples/aco_tsp/app.py
EwoutH: Marking this as resolved, but maybe something to keep in the back of our heads.

examples/conways_game_of_life_fast/app.py
```diff
@@ -20,11 +20,15 @@
     },
 }
 
+gol = GameOfLifeModel(10, 10)
+
+AlivePlot = make_plot_measure("Cells alive", "Fraction alive")
```
EwoutH, line 25: Does this plot nicely on a single plot? Since `"Cells alive"` might be on the 0-100 scale, while `"Fraction alive"` is in the 0-1 scale.

(they are a bit duplicate though, so removing one is also fine)

examples/hotelling_law/app.py
```diff
@@ -108,7 +108,8 @@ def agent_portrayal(agent):
     return portrayal
 
 
-def space_drawer(model, agent_portrayal):
+@solara.component
+def SpaceDrawer(model):
```
EwoutH, line 112: This function is capitalized like a Class. For my own knowlegde, does `@solara.component` make it Class-like?

EwoutH (approved):
Great work!

I especially like that the `agent_portrayal` now is a part of a the SpaceDrawer class/component, and you thus can have a portrayal for each drawer. Way more scalable!

examples/hotelling_law/app.py
Corvince: Aren't they all just objects? ;)

As far as I understood this is mostly a convention. It's best to imagine them as neither a class nor a function but as a solara component.

Corvince: Just to clarify, since I have looked it up by now. @solara.component does indeed return a regular ´class` object (with a a defined __call__ function, so you can call it like a function). Usually they should still be written in snake_case for standard Python formatting, but thats where the solara convention takes over

examples/conways_game_of_life_fast/app.py
Corvince: not really, I split this into 2 plots

PR #195, opened 2024-09-05
update boltzmann-wealth update to use new solaraviz api

Corvince:
Update to showcase the use of the new solaraviz API.

### DIFF
--- a/examples/boltzmann_wealth_model_experimental/app.py
+++ b/examples/boltzmann_wealth_model_experimental/app.py
@@ -1,4 +1,4 @@
-from mesa.visualization import SolaraViz
+from mesa.visualization import SolaraViz, make_plot_measure, make_space_matplotlib
 from model import BoltzmannWealthModel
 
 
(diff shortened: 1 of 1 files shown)


examples/boltzmann_wealth_model_experimental/app.py
```diff
@@ -24,11 +24,34 @@ def agent_portrayal(agent):
     "height": 10,
 }
 
+# Create initial model instance
+model1 = BoltzmannWealthModel(50, 10, 10)
+
+# Create visualization elements
+SpaceGraph = make_space_matplotlib(agent_portrayal)
+GiniPlot = make_plot_measure("Gini")
```
EwoutH, line 32: Maybe add a comment explaining that and why these are Classes,

EwoutH (approved):
Awesome, thanks!

examples/boltzmann_wealth_model_experimental/app.py
EwoutH: Thanks!

PR #202, opened 2024-09-21
Replace the remaining schedulers with AgentSet functionality

EwoutH:
This PR completes the migration from schedulers to AgentSet functionality across the mesa-examples repository for all regular (non-`gis`/-`rl`) examples. Key changes include:

- Replaced `RandomActivation`, `SimultaneousActivation`, and `RandomActivationByType` schedulers with appropriate AgentSet methods
- Updated `Model.step()` implementations to use AgentSet activation
- Removed references to `schedule.steps`, `schedule.agents`, and `schedule.agents_by_type`
- Updated agent addition/removal logic to work with AgentSets
- Adjusted data collection and visualization code to use `Model.steps` and `Model.agents`

For more details on migrating from schedulers to AgentSets, see the [migration guide](https://mesa.readthedocs.io/en/latest/migration_guide.html#time-and-schedulers).

PR #183, opened 2024-08-30
Reinstate PR #161: Replace schedulers with AgentSet functionality

EwoutH:
This reinstates PR #161 after the previous revert in #170.

Now that https://github.com/projectmesa/mesa/pull/2223 is merged, these examples are fully functional again without schedulers.

Original description of #161 below.

<hr>

This PR replace all the `RandomActivation` and `SimultaneousActivation` schedulers used in the example models with AgentSet functionality. This simplifies models and allows for easier and more flexible modification of these models. It algins the examples with the current best practices in Mesa.

### RandomActivation
The RandomActivation schedulers were removed, and in the Model.step() function the schedule step was replaced with an AgentSet operation, like:

```diff
-        self.schedule.step()
+        self.agents.shuffle().do("step")
```

The following RandomActivation schedulers were removed:
- bank_reserves
- boid_flockers
- boltzmann_wealth_model
- boltzmann_wealth_model_experimental
- boltzmann_wealth_model_network
- caching_and_replay
- charts
- el_farol
- epstein_civil_violence
- forest_fire
- hotelling_law
- schelling
- shape_example
- virus_on_network

### SimultaneousActivation
To remove the SimultaneousActivation schedulers, the Agent methods were renamed from `step()` and `advance()` to more descriptive functions. The model step was updated accordingly:

```diff
-        self.schedule.step()
+        self.agents.do("determine_state")
+        self.agents.do("assume_state")
```

The following SimultaneousActivation schedulers were removed:
- hex_snowflake
- color_patches
- conways_game_of_life

**Relevant tickets**
- https://github.com/projectmesa/mesa-examples/issues/111
- https://github.com/projectmesa/mesa/discussions/1912
- https://github.com/projectmesa/mesa/pull/1916
- https://github.com/projectmesa/mesa/pull/2223

PR #207, opened 2024-09-25
gis: Replace schedulers with AgentSet functionality

EwoutH:
Replace the schedulers in the GIS models with AgentSet functionality.

Similar to #183 and #202.

### DIFF
--- a/gis/agents_and_networks/src/model/model.py
+++ b/gis/agents_and_networks/src/model/model.py
@@ -22,7 +22,9 @@ def get_time(model) -> pd.Timedelta:
 
 def get_num_commuters_by_status(model, status: str) -> int:
     commuters = [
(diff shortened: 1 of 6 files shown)


gis/agents_and_networks/src/model/model.py
```diff
@@ -22,19 +22,21 @@ def get_time(model) -> pd.Timedelta:
 
 def get_num_commuters_by_status(model, status: str) -> int:
     commuters = [
-        commuter for commuter in model.schedule.agents if commuter.status == status
+        commuter
+        for commuter in model.agents_by_type[Commuter]
```
rht, line 26: Why is this `model.agents_by_type` instead of just `model.agents`?

EwoutH: Because there are also `Building` agents in the model.

PR #210, opened 2024-09-25
gis/GeoSir: Replace BaseScheduler with AgentSet functionality

### LINKED ISSUE #209: Replace GeoSir scheduler with AgentSet functionality
Currently the GeoSir model has interesting scheduling behavior, where it looks like the Agents are added multiple times to a base schedulers, with all Agents first being added, and then their neighbours again:

https://github.com/projectmesa/mesa-examples/blob/2c05149b4b52b70de77f4dfcbc59f4e3f93c554a/gis/geo_sir/model.py#L32
https://github.com/projectmesa/mesa-examples/blob/2c05149b4b52b70de77f4dfcbc59f4e3f93c554a/gis/geo_sir/model.py#L84-L90
https://github.com/projectmesa/mesa-examples/blob/2c05149b4b52b70de77f4dfcbc59f4e3f93c554a/gis/geo_sir/model.py#L107

The schedulers are deprecated (see https://github.com/projectmesa/mesa/pull/2306), so we would like to replace the BaseScheduler with `AgentSet.do()` operations (as of the [Migration guide](https://mesa.readthedocs.io/latest/migration_guide.html#time-and-schedulers)).

@glicerico, if I'm correct you originally wrote this model. Could you explain exactly what's happening? And would you be interested in updating the example?

EwoutH:
Replace BaseScheduler in the GeoSir gis model with AgentSet functionality, using `agents_by_type`, `shuffle_do()` and `do()`.

- Remove BaseScheduler initialization and usage
- Use automatic agent registration for PersonAgents
- Explicitly register NeighbourhoodAgents with the model
- Update step() method to use AgentSet methods for agent activation
- Maintain original activation order: PersonAgents (shuffled) then NeighbourhoodAgents
- Remove unnecessary scheduler.add() calls

Closes #209, follows up #207, motivated by https://github.com/projectmesa/mesa/pull/2306, similar to https://github.com/projectmesa/mesa-examples/pull/183 and https://github.com/projectmesa/mesa-examples/pull/202.

CC @glicerico.

### DIFF
--- a/gis/geo_sir/model.py
+++ b/gis/geo_sir/model.py
@@ -96,8 +96,9 @@ class GeoSir(mesa.Model):
         """Run one step of the model."""
         self.reset_counts()
 
-        # Activate PersonAgents first, in random order, then activate NeighbourhoodAgents.
+        # Activate PersonAgents in random order
         self.agents_by_type[PersonAgent].shuffle_do("step")
+        # For NeighbourhoodAgents the order doesn't matter, since they update independently from each other
         self.agents_by_type[NeighbourhoodAgent].do("step")
 
         self.datacollector.collect(self)
--- a/gis/geo_sir/model.py
+++ b/gis/geo_sir/model.py
@@ -29,7 +29,6 @@ class GeoSir(mesa.Model):
                                     if it has been exposed to another infected
         """
         super().__init__()
-        self.schedule = mesa.time.BaseScheduler(self)
         self.space = mg.GeoSpace(warn_crs_conversion=False)
         self.steps = 0
         self.counts = None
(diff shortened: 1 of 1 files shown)


gis/geo_sir/model.py
```diff
@@ -102,9 +94,11 @@ def reset_counts(self):
 
     def step(self):
         """Run one step of the model."""
-        self.steps += 1
         self.reset_counts()
-        self.schedule.step()
+
+        # Activate PersonAgents first, in random order, then activate NeighbourhoodAgents.
+        self.agents_by_type[PersonAgent].shuffle_do("step")
+        self.agents_by_type[NeighbourhoodAgent].do("step")
```
rht, line 101: Why does the ordering not matter here? Anyone reading the code would have the same question.

EwoutH: NeighbourhoodAgents work independently from each other. I will add a comment.

(note it’s already way more transparent then with the scheduler)

PR #215, opened 2024-10-03
Test SolaraViz in CI

EwoutH:
Initial attempt at testing the visualisation in CI.

@rht I can't get the imports working properly. I tried a lot of things, but somehow it always end up with one or more import errors. I know you're good at this, could you take a look?

### DIFF
--- a/.github/workflows/test_viz.yml
+++ b/.github/workflows/test_viz.yml
@@ -1,0 +1,61 @@
+name: Test viz
+
+on:
+  push:
+    paths:
+      - 'examples/**/app.py'  # If an example visualisation is modified
+      - 'examples/**/viz.py'  # If an example visualisation is modified
+      - 'viz.py'  # If the test script is modified
+      - '.github/workflows/test_viz.yml'  # If this workflow is modified
+  pull_request:
+    paths:
+      - 'examples/**/*.py'
+      - 'test_examples.py'
+      - '.github/workflows/test_examples.yml'
+  workflow_dispatch:
+  schedule:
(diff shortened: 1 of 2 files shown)


EwoutH:
Sorry pushed the wrong version, this is my attempt that got closest. It currently fails on imports in the models:

```
    import networkx as nx
    import solara
>   from aco_tsp.model import AcoTspModel, TSPGraph
E   ModuleNotFoundError: No module named 'aco_tsp'
```


```
    from mesa.visualization import SolaraViz, make_plot_measure
>   from model import GameOfLifeModel
E   ModuleNotFoundError: No module named 'model'

examples\conways_game_of_life_fast\app.py:2: ModuleNotFoundError
```

Sahil-Chhoker:
Hey @EwoutH, this file is written greatly. I was able to run step tests for the examples over on the main repository after a little tweaking.
But I wanted to ask how can something like `model.step()` go wrong that we need tests for it? Are there any cases where everything seems fine but `model.step()` gives error?

EwoutH:
Thanks, I think conceptual my approach is very sound. Unfortunately, I tried a million versions and never got it to work.

> I was able to run step tests for the examples over on the main repository after a little tweaking.

That's awesome!

> But I wanted to ask how can something like `model.step()` go wrong that we need tests for it? Are there any cases where everything seems fine but `model.step()` gives error?

Some of the visualisation logic only gets ran / evaluated when actually performing a model step (I think). So in this case we should be testing if the visualisation updates correctly after a model step, not the model step itself.

jackiekazil:
closing - stale.

PR #218, opened 2024-10-08
Updated mesa project

MariiaMelnychenko:
- У код було додано параметр randomize_new_cells, який дозволяє користувачеві вручну регулювати ймовірність випадкового оживлення нових клітин на кожному кроці моделювання. Цей параметр можна налаштувати через інтерфейс користувача, використовуючи повзунок. Це дає змогу контролювати динаміку моделі й експериментувати з частотою появи нових клітин під час симуляції.

- Також було додано нову логіку в правила гри: якщо клітина має більше ніж 4 живих сусіди, вона "вмирає" від перенаселення. Це додає більш реалістичної поведінки в гру, адже клітини можуть загинути через перенасичення живими сусідами. Клітини, що мають 2 або 3 сусіди, залишаються живими, а мертві клітини з рівно 3 сусідами оживають.

### DIFF
--- a/examples/conways_game_of_life_fast/app.py
+++ b/examples/conways_game_of_life_fast/app.py
@@ -34,9 +34,24 @@ model_params = {
         "max": 1,
         "step": 0.01,
     },
+    "randomize_new_cells": { #Новий параметр для ручного регулювання
+        "type": "SliderFloat",
+        "value": 0.05,  # Початкова ймовірність оживлення нових клітин
+        "label": "New Cells Randomization",
+        "min": 0,
+        "max": 1,
+        "step": 0.01,
+    },
 }
 
-gol = GameOfLifeModel()
+
+gol = GameOfLifeModel(
+    width=model_params["width"]["value"],
+    height=model_params["height"]["value"],
+    alive_fraction=model_params["alive_fraction"]["value"],
+    randomize_new_cells=model_params["randomize_new_cells"]["value"]
+)
+
 
 layer_viz = make_space_matplotlib(propertylayer_portrayal=propertylayer_portrayal)
 TotalAlivePlot = make_plot_measure("Cells alive")
--- a/examples/conways_game_of_life_fast/model.py
+++ b/examples/conways_game_of_life_fast/model.py
@@ -7,7 +7,7 @@ from scipy.signal import convolve2d
 
 # fmt: off
 class GameOfLifeModel(Model):
-    def __init__(self, width=10, height=10, alive_fraction=0.2):
+    def __init__(self, width=10, height=10, alive_fraction=0.2, randomize_new_cells=0.05):
         super().__init__()
         # Initialize the property layer for cell states
         self.cell_layer = PropertyLayer("cells", width, height, False, dtype=bool)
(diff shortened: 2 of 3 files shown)


EwoutH:
Thanks for opening a PR! Could you translate the Ukrainian part to English?

I’m not sure if I want to alter the classic Game of Life rules themselves, but the “noise” concept could be interesting.

EwoutH:
Thanks for translating! The PR is appreciated.

However, currently it changes too much unnecessarily. I'm also not fully convinced of changing such a classic, well defined model, like [Conway's Game of Life](https://en.wikipedia.org/wiki/Conway%27s_Game_of_Life).

The concept of noise/mutations is interesting though. I think you can go two ways with this:

- "random mutations" (what you currently implemented): Each Agent can have a certain chance for a certain property to change each timestep
- "communication noise": There's a chance that either information is lost or the decision is make incorrectly when an action if performed.
  - A neighbour cell is read incorrectly
  - The rule isn't applied (or inversed)

These are interesting concepts. Only maybe is this not the model to apply it on.

@projectmesa/maintainers would appreciate some other opinions, especially on in which models this might be useful (if at all).

tpike3:
> Thanks for translating! The PR is appreciated.
> 
> However, currently it changes too much unnecessarily. I'm also not fully convinced of changing such a classic, well defined model, like [Conway's Game of Life](https://en.wikipedia.org/wiki/Conway%27s_Game_of_Life).
> 
> The concept of noise/mutations is interesting though. I think you can go two ways with this:
> 
> * "random mutations" (what you currently implemented): Each Agent can have a certain chance for a certain property to change each timestep
> * "communication noise": There's a chance that either information is lost or the decision is make incorrectly when an action if performed.
>   
>   * A neighbour cell is read incorrectly
>   * The rule isn't applied (or inversed)
> 
> These are interesting concepts. Only maybe is this not the model to apply it on.
> 
> @projectmesa/maintainers would appreciate some other opinions, especially on in which models this might be useful (if at all).
> 
> @MariiaMelnychenko I would love to hear the background what you're using Mesa for, as part of which project/course/interest. Might help us help you better :)

As they are parameters, I think we should allow it. This can allow folks to explore the model and it underlying concepts more. 

So I would say it defaults to the Conways classical model, but now has sliders/paramters to allow users to explore different variants.

rht:
You can always fork the classic model into a separate, adjacent model in another folder. Adding variations to the classic model will strain bandwidth of the learner who is trying to understand the essence of the model.

In short, I think it's better to create a separate model, named "conways_game_of_life_advanced", where any variations and extensions can be added.

quaquel:
Since I have been going through all the examples for the grid spaces, I feel that we have too many examples and variants of essentially the same model. 

I recall a discussion on this before, but my idea would be to have a carefully curated set of famous examples in the main mesa repo (i.e., wolf-sheep, Schelling, Epstein civil violence, sugarscape, Axelrod evolution of collaboration, el farol, forest fire, Conway game of life, and virus on a network), and have a not actively maintained gallery for show and tell. We could do a template for the gallery that includes an environment file to enhance future reproducibility. The curated examples would be used for integration testing and be kept in sync with the main branch at all times.

jackiekazil:
Conceptually - there is the possibility of an infinite number of models that can be created off of one model.
I like what is being submitted here. During today's dev meeting, we discussed moving the core examples back to the main repo -- what if we created a Conway's folder with a readme that pointed to the core example (when it is moved), but also included this as an alternate example, which highlights X. One day if we have 10 versions -- the readme also lists each version and why you might want to look or consider it. Think of it like a writing assignment for 10 people come up with 10 different approaches. Thoughts?

PR #198, opened 2024-09-10
Moving examples to use the new discrete spaces

quaquel:
This PR moves over all relevant examples to the experimental grid spaces. While making this move, the experimental grid spaces continue to be refined.

PR #222, opened 2024-10-13
refactor: Simplify Schelling code 

rht:
1. Remove unused model attributes
2. Make `similar` calculation more natural language readable

With these changes and `shuffle_do`, the Schelling code in https://github.com/JuliaDynamics/ABMFrameworksComparison should go down from 44 to ~33, closer to Agents.jl's 28.

### DIFF
--- a/examples/schelling/model.py
+++ b/examples/schelling/model.py
@@ -88,5 +88,4 @@ class Schelling(mesa.Model):
 
         self.datacollector.collect(self)
 
-        if self.happy == len(self.agents):
-            self.running = False
+        self.running = self.happy != len(self.agents)
--- a/examples/schelling/model.py
+++ b/examples/schelling/model.py
@@ -18,12 +18,10 @@ class SchellingAgent(mesa.Agent):
         self.type = agent_type
 
     def step(self):
-        similar = 0
-        for neighbor in self.model.grid.iter_neighbors(
+        neighbors = self.model.grid.iter_neighbors(
             self.pos, moore=True, radius=self.model.radius
-        ):
-            if neighbor.type == self.type:
-                similar += 1
+        )
+        similar = sum(1 for neighbor in neighbors if neighbor.type == self.type)
 
         # If unhappy, move:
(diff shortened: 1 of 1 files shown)


examples/schelling/model.py
```diff
@@ -6,24 +6,21 @@ class SchellingAgent(mesa.Agent):
     Schelling segregation agent
     """
 
-    def __init__(self, model, agent_type):
+    def __init__(self, model: mesa.Model, agent_type: int) -> None:
         """
         Create a new Schelling agent.
```
rht, line 11: This docstring is self explanatory for anyone knowing what `__init__` does, and should be considered a superfluous documentation.

rht: I have read several articles about superfluous documentation, but couldn't find a good reference that explains it. Maybe this [SO thread](https://stackoverflow.com/questions/33019955/should-you-always-document-functions-even-if-redundant-specifically-python) does it.

examples/schelling/model.py
```diff
@@ -95,5 +88,4 @@ def step(self):
 
         self.datacollector.collect(self)
 
-        if self.happy == len(self.agents):
-            self.running = False
+        self.running = self.happy != len(self.agents)
```
EwoutH, line 91: While this is fully correct, I find it less readable. Maybe this could also be resolved with a comment.

EwoutH (approved):
Thanks, especially the step simplification is really nice.

One comment, otherwise good. Maybe check with @quaquel if it doesn’t conflict his work.

quaquel:
It does conflict with #198 but I can easily rectify that there once this is merged. 

It would be good to move these changes also to the benchmarks in the main repo (is this an argument in favor of [#2349](https://github.com/projectmesa/mesa/pull/2349)? 😉 )

examples/schelling/model.py
quaquel: My experience is the other way around. I found this very clear.

quaquel (approved):

PR #223, opened 2024-10-15
bug fix and code simplification in hotelling

quaquel:
follow up from #198

### DIFF
--- a/examples/hotelling_law/hotelling_law/agents.py
+++ b/examples/hotelling_law/hotelling_law/agents.py
@@ -9,7 +9,7 @@ class StoreAgent(CellAgent):
     """An agent representing a store with a price and ability to move
     and adjust prices."""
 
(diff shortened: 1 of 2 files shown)


EwoutH (approved):
Awesome, appreciated, and really clean!

That `all_cells.select_random_cell()` is just such a handy construct to have!

Maybe we can even add `random_cell` and `random_empty_cell` properties to the spaces, so that you don't have to go via `all_cells` (just an idea, other PR of course).

quaquel:
> Maybe we can even add random_cell and random_empty_cell properties to the spaces, so that you don't have to go via all_cells (just an idea, other PR of course).

Yes having it on the grid would be easy (as would be a method for drawing 1 cells with replacement and without replacement - Both can probably be added to `CellCollection first).

PR #235, opened 2024-11-11
updates examples to explicitly set ranom on new style grid spaces

quaquel:
[#2479](https://github.com/projectmesa/mesa/pull/2479) results in user warnings if random is not passed explicitly in `DiscreteSpace`, `AgentSet`, and `CellCollection`. This fixes it for the examples that did not do this correctly.


EwoutH (approved):
Thanks, appreciated!

PR #241, opened 2024-12-27
added a new example: city walking behaviour

Sahil-Chhoker:
Draft PR for a new example in mesa-examples repository.

### DIFF
--- a/examples/city_walking_behaviour/Readme.md
+++ b/examples/city_walking_behaviour/Readme.md
@@ -1,0 +1,24 @@
+# Walking Behavior Agent-Based Model
+
+This repository contains an agent-based model (ABM) that simulates walking behavior in a hypothetical city, examining how socioeconomic status (SES), built environment, and social factors influence walking patterns.
+
+This ABM investigates spatial determinants of social inequalities in walking behavior by modeling dynamic interactions between individual attributes and environmental factors. The model incorporates feedback mechanisms and individual-environment interactions to simulate realistic walking patterns across different socioeconomic groups.
+
+## How to Run
+
+To run a basic simulation:
+
+```python
+
+solara run app.py
+
+```
+
+# Files
+
+- [city_walking_behaviour/model.py](city_walking_behaviour/model.py): Core model file.
+- [city_walking_behaviour/agents.py](city_walking_behaviour/agents.py): The agent class.
+
+## Further Reading
+
+1. A Spatial Agent-Based Model for the Simulation of Adults’ Daily Walking Within a City [article](https://pmc.ncbi.nlm.nih.gov/articles/PMC3306662/)
--- a/examples/city_walking_behaviour/city_walking_behaviour/agents.py
+++ b/examples/city_walking_behaviour/city_walking_behaviour/agents.py
@@ -1,0 +1,413 @@
+from typing import Optional
+from mesa import Model
+import math
+from mesa.agent import AgentSet
+from mesa.experimental.cell_space import CellAgent, FixedAgent
+from mesa.experimental.cell_space import Cell
+from enum import Enum
+import numpy as np
+
+# Constants for probability values
+FEMALE_PROBABILITY = 0.5
+DOG_OWNER_PROBABILITY = 0.2
+SINGLE_HOUSEHOLD_PROBABILITY = 0.2
+MIN_AGE = 18
+MAX_AGE = 87
+MIN_FRIENDS = 3
+MAX_FRIENDS = 5
+WORKING_PROBABILITY = 0.95
+RETIREMENT_AGE = 69
+
+
+class ActivityType(Enum):
+    WORK = "work"
+    GROCERY = "grocery"
+    NON_FOOD_SHOPPING = "shopping"
+    SOCIAL = "social"
+    LEISURE = "leisure"
+
+
+class Workplace:
+    """Generic workplace base class."""
+
+    def __init__(self, store_type: str):
+        self.store_type = store_type
+
+
+class GroceryStore(Workplace, FixedAgent):
+    def __init__(self, model: Model, cell=None):
+        Workplace.__init__(self, store_type="Grocery Store")
+        FixedAgent.__init__(self, model)
+        self.cell = cell
+
+
+class NonFoodShop(Workplace, FixedAgent):
+    def __init__(self, model: Model, cell=None):
+        Workplace.__init__(self, store_type="Non-Food Shop")
+        FixedAgent.__init__(self, model)
+        self.cell = cell
+
+
+class SocialPlace(Workplace, FixedAgent):
+    def __init__(self, model: Model, cell=None):
+        Workplace.__init__(self, store_type="Social Place")
+        FixedAgent.__init__(self, model)
+        self.cell = cell
+
+
+class Other(Workplace, FixedAgent):
+    def __init__(self, model: Model, cell=None):
+        Workplace.__init__(self, store_type="Other")
+        FixedAgent.__init__(self, model)
+        self.cell = cell
+
+
+class WalkingBehaviorModel:
+    """Encapsulates all walking decisions for a human agent, including distance checks."""
+
+    DAILY_PROBABILITIES = {
+        ActivityType.GROCERY: 0.25,  # Every 4 days on average
+        ActivityType.NON_FOOD_SHOPPING: 0.25,
+        ActivityType.SOCIAL: 0.15,
+        ActivityType.LEISURE: 0.20,
+    }
+
+    BASE_MAX_DISTANCES = {
+        ActivityType.WORK: 2000,  # meters
+        ActivityType.GROCERY: 1000,
+        ActivityType.NON_FOOD_SHOPPING: 1500,
+        ActivityType.SOCIAL: 2000,
+        ActivityType.LEISURE: 3000,
+    }
+
+    def __init__(self, model: Model):
+        self.model = model
+        self.total_distance_walked = 0
+
+    def reset_daily_distance(self):
+        """Reset the total distance walked for a new day"""
+        self.total_distance_walked = 0
+
+    def add_distance(self, distance: float):
+        """Add distance to total daily walking distance"""
+        self.total_distance_walked += distance
+
+    def get_max_walking_distance(self, human, activity: ActivityType) -> float:
+        """Calculate person-specific maximum walking distance"""
+        return self.BASE_MAX_DISTANCES[activity] * human.walking_ability
+
+    def calculate_distance(self, cell1, cell2) -> float:
+        """Calculate distance between two cells"""
+        x1, y1 = cell1.coordinate
+        x2, y2 = cell2.coordinate
+        return math.sqrt((x1 - x2) ** 2 + (y1 - y2) ** 2)
+
+    def decide_walk_to_work(self, human) -> bool:
+        """Decide if person walks to work"""
+        if not human.is_working or not human.workplace:
+            return False
+
+        distance = self.calculate_distance(human.cell, human.workplace.cell)
+        max_distance = self.get_max_walking_distance(human, ActivityType.WORK)
+
+        if distance <= max_distance:
+            return self.model.random.random() <= human.walking_attitude
+        return False
+
+    def find_nearest_location(
+        self, human, activity_type: ActivityType, search_workplace: bool = True
+    ) -> Optional[FixedAgent]:
+        """Find nearest location of given type within walking distance"""
+        if activity_type == ActivityType.GROCERY:
+            locations = self.model.agents_by_type[GroceryStore]
+        elif activity_type == ActivityType.NON_FOOD_SHOPPING:
+            locations = self.model.agents_by_type[NonFoodShop]
+        elif activity_type == ActivityType.SOCIAL:
+            locations = self.model.agents_by_type[SocialPlace]
+        else:
+            return None
+
+        max_distance = self.get_max_walking_distance(human, activity_type)
+
+        # Check locations near home
+        nearest = min(
+            locations,
+            key=lambda loc: self.calculate_distance(human.cell, loc.cell),
+            default=None,
+        )
+        if (
+            nearest
+            and self.calculate_distance(human.cell, nearest.cell) <= max_distance
+        ):
+            return nearest
+
+        # Check locations near workplace if applicable
+        if search_workplace and human.workplace:
+            nearest = min(
+                locations,
+                key=lambda loc: self.calculate_distance(human.workplace.cell, loc.cell),
+                default=None,
+            )
+            if (
+                nearest
+                and self.calculate_distance(human.workplace.cell, nearest.cell)
+                <= max_distance
+            ):
+                return nearest
+
+        return None
+
+    def decide_leisure_walk(self, human) -> Optional[tuple[float, float]]:
+        """Decide if person takes a leisure walk"""
+        if (
+            self.model.random.random()
+            > self.DAILY_PROBABILITIES[ActivityType.LEISURE] * human.walking_attitude
+        ):
+            return None
+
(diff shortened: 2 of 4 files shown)


EwoutH:
Looks like quite a serious model!

If you could add to the PR description:
- A summary / overview of the model
- Any outstanding questions / considerations

Sahil-Chhoker:
1. I have updated the Readme file to contain the basic functionality of the model and agents.
2. Some shortcomings:
- Very slow.
- Unable to simulate with recommended grid size (800x800) and human population.
- Wrong results noted:
- Screenshot 2024-12-26 161405
   to be compared with
- accumulated_graphs
(attention to be paid to the SES color being different in both the graphs)

EwoutH:
Thanks for the extended descriptions.

Have you profiled the model to identify the main performance bottlenecks?

EwoutH:
@projectmesa/maintainers I’m a bit in doubt on the complexity of this example. On the one hand it shows you can create ABMs with detailed environments and complex behavior, on the other hand I don’t know an example this extended would really benefit our users. I would like to hear some of your stances.

Sahil-Chhoker:
> Have you profiled the model to identify the main performance bottlenecks?

I haven't, can you tell how can I do that?

EwoutH:
Some starting point:
- https://docs.python.org/3/library/profile.html
- https://realpython.com/python-profiling/

Most IDEs also have a profiler built-in.

Sahil-Chhoker:
Here is the table I built from the data collected by the profiler.
### Performance Analysis Results

| Function/Method Name          | NCalls  | TotTime | CumTime | %Total |
|------------------------------|---------|---------|---------|--------|
| step(agents.py)              |   4,850 |   0.019 |   0.596 |  74.4% |
| simulate_daily_walks         |   4,850 |   0.033 |   0.426 |  53.2% |
| decide_leisure_walk          |   4,850 |   0.092 |   0.226 |  28.2% |
| calculate_distance           | 185,574 |   0.093 |   0.132 |  16.5% |
| get_feedback                 |   1,303 |   0.041 |   0.128 |  16.0% |
| weakref.keys                 | 185,980 |   0.072 |   0.113 |  14.1% |
| find_nearest_location        |   2,391 |   0.010 |   0.109 |  13.6% |
| min (built-in)              |   5,689 |   0.018 |   0.093 |  11.6% |
| sum (built-in)              |  14,770 |   0.012 |   0.068 |   8.5% |
| lambda (agents.py:135)       |  17,825 |   0.018 |   0.041 |   5.1% |
| sqrt (built-in)             | 190,260 |   0.039 |   0.039 |   4.9% |
| cell (cell_agent.py)        | 188,561 |   0.029 |   0.029 |   3.6% |

EwoutH:
Okay, this is quite useful already! It shows that most of the runtime is taken up in the Agent `step`,  and specifically the decision logic.

You can try to speed that up, but if you want I can also give it a try (hopefully tomorrow or Monday).

Sahil-Chhoker:
> You can try to speed that up, but if you want I can also give it a try (hopefully tomorrow or Monday).

I'll do what I can, but I'm not particularly confident. Can we use caching to prevent duplicate lookups in `calculate_distance`? I'm also not sure what could speed up the other functions.

EwoutH:
Hypothesis for model results not correlating with the paper:

1. Safety and aesthetics are note implemented correctly (check by adding visualisation of PropertyLayers)
2. Agents are not reading the correct values from the PropertyLayers
3. Agents are not using the read value correctly in their behavior

EwoutH:
In theory this should work:

> I'll help you visualize both property layers by updating the space component in app.py. We'll create a portrayal dictionary for the property layers that defines a color gradient from transparent to solid blue/green.
> 
> Here's how to modify the app.py code. First, add this portrayal dictionary right after the model_params definition:
> 
> ```python
> propertylayer_portrayal = {
>     "safety": {
>         "color": "blue",
>         "alpha": 0.5,
>         "colorbar": True
>     },
>     "aesthetic": {
>         "color": "green", 
>         "alpha": 0.5,
>         "colorbar": True
>     }
> }
> ```
> 
> Then update the space_component definition to include the property layers:
> 
> ```python
> space_component = make_space_component(
>     agent_portrayal, 
>     draw_grid=True, 
>     post_process=post_process_space,
>     propertylayer_portrayal=propertylayer_portrayal
> )
> ```
> 
> This will:
> 1. Draw the safety layer in blue gradient where low values are transparent and high values are solid blue
> 2. Draw the aesthetic layer in green gradient where low values are transparent and high values are solid green
> 3. Display colorbars for both layers to show the value scale
> 4. Set alpha to 0.5 so the layers blend nicely and don't completely obscure agents
> 
> The layers will be drawn before the agents, so you'll still be able to see all the agents on top of the colored background created by these layers.

EwoutH:
- https://github.com/projectmesa/mesa/pull/2336

Sahil-Chhoker:
I have successfully visualized the property layers, and they are functioning as intended. Before moving on to the other points, I reread the paper and realized I had overlooked the concept of households. I will now implement the household concept and check the results again.

EwoutH:
Awesome, that's amazing to hear! Did you figure out what the bug was or what I missed yesterday?

EwoutH:
> I reread the paper and realized I had overlooked the concept of households. I will now implement the household concept and check the results again.

CC @tpike3, Households sounds like Meta agents. Maybe you could assist here!

- https://github.com/projectmesa/mesa/pull/2575

Sahil-Chhoker:
@EwoutH Unfortunately no, but since they are just a bunch of numpy arrays I exported the data and made some graphs based on the values. I can confirm they are implemented as they are intended.

> CC @tpike3, Households sounds like Meta agents. Maybe you could assist here!

I read a bit about MetaAgents, and from my understanding, they are groups of agents that form during the model runtime based on certain properties. While I can see their usefulness, I don’t think they are particularly relevant to this model. This is because the household system is defined during the human placement, and a household is simply the cell where they spawn. Therefore, in my opinion, this can be handled during initialization.

EwoutH:
Thanks, sounds good!

Sahil-Chhoker:
@EwoutH I have a question: is it correct to use `model.random` everywhere inside the agents' logic? For example, if agents need to find new locations to move to every day, wouldn’t using the same `model.random` result in the agents walking to the same place they walked the previous day?

quaquel:
it's better to use `self.random` inside an `Agent`, so `self.random.random()` will draw a random number of unit interval. In either case `agent.random` or `model.random` just refer to the same seeded python stdlib `random.Random` instance which is a random number generator so repeated calls will return different numbers.

Sahil-Chhoker:
The model has improved significantly with the following changes:  
1. Rewriting the walking behavior to incorporate household positions.  
2. Improving the placement of agents based on their SES.  
3. Scaling the feedback factor by 20 due to its current size.  

**Tasks to be completed:**  
1. Fix the logic for leisure walks.  
2. Review the positioning of agents.  

current results for RR:
Screenshot 2024-12-31 184428
observations:
1. It becomes stable like in the graph from paper.
2. Distribution by SES is correct.

Sahil-Chhoker:
I have thoroughly reviewed the model in every aspect and believe I have covered everything correctly. In my opinion, the only remaining task is to make the model efficient enough to run on an 800x800 grid or decide to keep this simpler model after applying smaller optimizations.

Current Results:
image


Current Profile: Running the model with 20x20 grid and 1350 humans for 10 steps.
<html>
<body>

Function | ncalls | tottime | percall | cumtime | percall
-- | -- | -- | -- | -- | --
_wrapped_step | 10 | 0.000 | 0.000 | 2.300 | 0.230
step | 10 | 0.000 | 0.000 | 2.300 | 0.230
shuffle_do | 10 | 0.023 | 0.002 | 2.165 | 0.217
step (agents) | 13,500 | 0.055 | 0.000 | 2.119 | 0.000
simulate_daily_walks | 13,500 | 0.101 | 0.000 | 1.636 | 0.000
decide_leisure_walk | 13,500 | 0.024 | 0.000 | 0.920 | 0.000
get_leisure_cells | 1,208 | 0.565 | 0.000 | 0.885 | 0.001
__init__ (model) | 1 | 0.000 | 0.000 | 0.761 | 0.761
add_initial_humans | 1 | 0.007 | 0.007 | 0.715 | 0.715
__init__ (agents) | 1,350 | 0.011 | 0.000 | 0.687 | 0.001
get_workplace | 1,350 | 0.245 | 0.000 | 0.561 | 0.000
keys | 1,217,553 | 0.348 | 0.000 | 0.415 | 0.000
find_walkable_locations | 8,851 | 0.038 | 0.000 | 0.404 | 0.000
get_feedback | 3,001 | 0.133 | 0.000 | 0.369 | 0.000
check_near_point | 10,022 | 0.126 | 0.000 | 0.254 | 0.000
sum | 49,697 | 0.038 | 0.000 | 0.223 | 0.000
calculate_distance | 368,985 | 0.137 | 0.000 | 0.212 | 0.000
collect | 11 | 0.001 | 0.000 | 0.147 | 0.013
abs | 864,370 | 0.125 | 0.000 | 0.125 | 0.000
mean | 3,001 | 0.051 | 0.000 | 0.117 | 0.000
get_distance | 87,277 | 0.083 | 0.000 | 0.103 | 0.000


</body>
</html>

examples/city_walking_behaviour/Readme.md
EwoutH, line 1: It might be nice to add a high-level overview of the (expected) results in this Readme, and maybe a screenshot or two to support that.

EwoutH:
Awesome, I will do a full review this weekend.

examples/city_walking_behaviour/Readme.md
Sahil-Chhoker: Understood!

Sahil-Chhoker:
@EwoutH I would love to hear your thoughts on this.

examples/city_walking_behaviour/city_walking_behaviour/agents.py
```diff
@@ -0,0 +1,526 @@
...
+
+
+class Other(Workplace, FixedAgent):
+    def __init__(self, model: Model, cell=None):
+        Workplace.__init__(self, store_type="Other")
+        FixedAgent.__init__(self, model)
+        self.cell = cell
+
+
+class WalkingBehaviorModel:
```
EwoutH, line 67: Maybe rename this to avoid confusion with the Mesa model.

examples/city_walking_behaviour/city_walking_behaviour/model.py
```diff
@@ -0,0 +1,729 @@
...
+        width: int = 40,
+        no_of_couples: int = 2400,
+        no_of_singles: int = 600,
+        no_of_grocery_stores: int = 10,
+        no_of_social_places: int = 75,
+        no_of_non_food_shops: int = 40,
+        no_of_others: int = 475,
+        scenario: str = "random_random",
+        seed=None,
+        simulator=ABMSimulator(),
```
EwoutH, line 43: Are you actually using DEVS functionality? I can't find it in your code, so I'm curious if you use the functionality.

examples/city_walking_behaviour/city_walking_behaviour/model.py
```diff
@@ -0,0 +1,729 @@
...
+                    for agent in x.agents_by_type[Human]
+                    if agent.SES == 5
+                )
+                / len(x.agents_by_type[Human])
+                if x.agents_by_type[Human]
+                else 0
+            ),
+        }
+
+        self.datacollector = DataCollector(model_reporters)
```
EwoutH, line 271: Maybe we can generate these a bit more elegantly, can you check if something like this works?
```suggestion
        def create_model_reporters():
            """Create a dictionary of model reporters with minimal boilerplate"""
            reporters = {}
            metrics = ['daily_walking_trips', 'work_trips', 'basic_needs_trips', 'leisure_trips']
            
            for metric in metrics:
                for ses in range(1, 6):
                    name = f"avg_{metric.split('_')[0]}_ses{ses}"  # e.g. "avg_walk_ses1"
                    
                    # Create the reporter function
                    reporters[name] = lambda x, m=metric, s=ses: (
                        sum(getattr(agent, m) for agent in x.agents_by_type[Human] if agent.SES == s)
                        / len(x.agents_by_type[Human])
                        if x.agents_by_type[Human]
                        else 0
                    )
            
            return reporters

        # Generate the model_reporters dict
        model_reporters = create_model_reporters()

        self.datacollector = DataCollector(model_reporters)
```

examples/city_walking_behaviour/city_walking_behaviour/model.py
```diff
@@ -0,0 +1,729 @@
...
+
+    def step(self):
+        """Advance the model by one step."""
+        self.agents_by_type[Human].shuffle_do("step")
+
+        # Reset daily walking trips
+        self.agents_by_type[Human].daily_walking_trips = 0
+        self.agents_by_type[Human].work_trips = 0
+        self.agents_by_type[Human].basic_needs_trips = 0
+        self.agents_by_type[Human].leisure_trips = 0
```
EwoutH, line 712: Maybe give `Human` a method `reset_daily_walking_trips` and call that one here.

examples/city_walking_behaviour/city_walking_behaviour/agents.py
```diff
@@ -0,0 +1,526 @@
...
+        Workplace.__init__(self, store_type="Social Place")
+        FixedAgent.__init__(self, model)
+        self.cell = cell
+
+
+class Other(Workplace, FixedAgent):
+    def __init__(self, model: Model, cell=None):
+        Workplace.__init__(self, store_type="Other")
+        FixedAgent.__init__(self, model)
+        self.cell = cell
```
EwoutH, line 64: I'm curious why these are classes, and not just a `Place` with a `place_type` variable.

It might be that there's a very good reason, just curious.

examples/city_walking_behaviour/city_walking_behaviour/agents.py
```diff
@@ -0,0 +1,526 @@
...
+        return math.sqrt(dx * dx + dy * dy)
+
+    def get_distance(self, cell1, cell2) -> float:
+        """Get distance between cells using cache."""
+        key = (cell1, cell2)
+        if key not in self._distance_cache:
+            x1, y1 = cell1.coordinate
+            x2, y2 = cell2.coordinate
+            self._distance_cache[key] = self.calculate_distance(x1, y1, x2, y2)
+        return self._distance_cache[key]
```
EwoutH, line 123: @quaquel this example model uses the cell space. Since it's one of the first user models using it, could you check if:

1. everything is used according to best practices
2. there are inconsistencies, missing function or opportunities to improve the cell space

examples/city_walking_behaviour/city_walking_behaviour/model.py
```diff
@@ -0,0 +1,729 @@
...
+                    SES=ses,
+                    unique_id=self.unique_id,
+                    cell=cell,
+                    household=household,
+                )
+                self.unique_id += 1
+
+    def step(self):
+        """Advance the model by one step."""
+        self.agents_by_type[Human].shuffle_do("step")
```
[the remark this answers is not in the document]
Sahil-Chhoker: They are not added or removed, therefore I made a `human_set` that keeps track of every `human` and calls `shuffle_do` on `human_set` instead.

examples/city_walking_behaviour/city_walking_behaviour/agents.py
Sahil-Chhoker: I personally prefer this separate classes design choice since it makes the codebase more extensible - if we later need to add specific behaviors for different place types (like closed during certain days), it'll be cleaner to add those to the relevant classes rather than adding conditional logic in a single class.

EwoutH: Sounds good, thanks

PR #252, opened 2025-04-14
Add New Preferential Attachment Example

Spartan-71:
### Summary

This PR introduces a new **Preferential Attachment** simulation example, inspired by the Barabási–Albert model of network growth. The simulation models how a network evolves over time as new nodes are added and preferentially attach to existing nodes with higher degrees, leading to the emergence of scale-free network structures.

---

### Features

- **Node Agents**: Represent individual nodes in the network.
- **Preferential Attachment Rule**: New nodes attach to existing nodes with probability proportional to node degree.
- **Data Collection**: Tracks the number of nodes with degree 1 over time.
- **Dynamic network topology**: Dynamic graph representation of the growing network.

---

### Implementation Details

- Uses Mesa’s `Network` grid to embed and manage the graph.
- Relies on `networkx` for graph data structure and operations.
- Probabilities for attachment are derived from current node degrees.
- Includes a `DataCollector` to track key metrics like node degrees.

---

### Visualization

image

---


EwoutH:
Thanks for another example! Very neat that you keep building things :)

@tpike3 and @quaquel: This example is using the new discrete network space, curious what you think!

@Sahil-Chhoker and/or @sanika-n could (one of) you review the visualization (implementation)?

@Spartan-71 I’m curious, when building these example buildings, were there any Mesa features you were missing, felt counterintuitive or could be improved?

Sahil-Chhoker:
I just looked at the NetLogo's implementation of this model, it looks absolutely beautiful! And looking at that I also want this model to have similar plots (degree distribution simple and with log) and infinite number of agents if its possible @Spartan-71.

Spartan-71:
Ohky will try to get the same in mesa by the end of the day.

examples/preferencial_attachment/preferencial_attachment/model.py
```diff
@@ -0,0 +1,60 @@
...
+        total_degree = sum(degree)
+
+        # probabilities of connecting to an existing node
+        probabilities = [d / total_degree for d in degree]
+
+        # choose an existing node based on the computed probabilities
+        chosen_node = np.random.choice(nodes, p=probabilities)
+
+        # add the new node and connect it to the choosen node
+        self.graph.add_edge(chosen_node, self.curr_node)
```
quaquel, line 56: This won't work. You merely add a link in the graph that is used as the basis for the Network space. However, this additional edge in the graph won't be reflected in the Network space. 

In fact, dynamic networks are not possible at the moment with discrete space because discrete spaces rely on caching their connections.

tpike3: @Spartan-71 With the release of 3.2 and @quaquel' work, we have the ability to have dynamic space if you want to take a look for this example.

PR #253, opened 2025-04-18
virus antibody model

colinfrisch:
# Virus-Antibody Model

This model is a simulation of immune reaction declined as a confrontation between antibody agents and virus agents. The global idea is to model how the imune system can struggle against new virus but is able to adapt over time and beat a same virus if it comes back. The results are quite interesting as the simulation can go both ways (virus win or antibodies win) with a little tweak in the base parameters (check the readme for some examples).

## What it showcases 
- **Usage of memory in agents** : divided into a short term memory using a deque to easily add and remove memories in case of a new virus encounter, and a long term memory (here a simple list)
- **Agent knowledge sharing** : the antibodies are able to share short term memory)
- **Usage of weak referencing** to avoid coding errors (antibodies can store viruses in a `self.target` attribute)
- Emergence of completely **different outcomes** with only small changes in parameters

## What it adds to mesa-examples 
- Will be the first continuous‑space user example
- Weak‑referencing and dual memory systems and share knowledge (cf. what it showcases)
- Theme is original and quite different from what already exists in examples : shows the diversity of fields in which mesa can be useful

## How It Works

1. **Initialization**: The model initializes a population of viruses and antibodies in a continuous 2D space.
2. **Agent Behavior**:
   - Antibodies move randomly until they detect a virus within their sight range (becomes purple), than pursue the virus.
   - Antibodies pass on all the virus DNA in their short term memory to the nearest antibodies (cf. example)
   - Viruses move randomly and can duplicate or mutate.
3. **Engagement (antibody vs virus)**: When an antibody encounters a virus:
   - If the antibody has the virus's DNA in its memory, it destroys the virus.
   - Otherwise, the virus may defeat the antibody, causing it to lose health or become inactive temporarily.
4. **Duplication**: Antibodies and viruses can duplicate according to their duplication rate.

pattern


examples/virus_antibody/agents.py
```diff
@@ -0,0 +1,259 @@
...
+        ko_timeout,
+        memory_capacity,
+        initial_position=(0, 0),
+        direction=(1, 1),
+    ):
+        super().__init__(model=model, space=space)
+
+        # Movement & state
+        self.position = initial_position
+        self.speed = 1.5
```
quaquel, line 37: Would it be possible to make speed and health kwargs on the init? In my view it is good practice to make all attributes explicitly controllable.

examples/virus_antibody/agents.py
```diff
@@ -0,0 +1,259 @@
...
+        # Then move
+        self.move()
+
+    def find_closest_virus(self):
+        agents, _ = self.space.get_agents_in_radius(self.position, self.sight_range)
+        viruses = [a for a in agents if isinstance(a, VirusAgent)]
+        return viruses[0] if viruses else None
+
+    def communicate(self) -> bool:
+        agents, _ = self.space.get_agents_in_radius(self.position, self.sight_range)
```
quaquel, line 76: You do the `space.get_agents_in radius` twice. Once in find_closest_virus, and once in communicate. It would be computationally more efficient to do it once in step, split the return in virus and antibody agents, and use these respectively in the helper methods. This will substantially reduce runtime.

examples/virus_antibody/agents.py
```diff
@@ -0,0 +1,259 @@
...
+            initial_position=self.position,
+            direction=self.direction,
+        )
+        # Copy over memory
+        clone.st_memory = deque(item for item in self.st_memory if item)
+        clone.lt_memory = [item for item in self.lt_memory if item]
+        clone.target = None
+        clone.ko_steps_left = 0
+
+        self.model.antibodies_set.add(clone)
```
quaquel, line 113: this should not be needed. `model.agents_by_type[AntibodyAgent]` will handle this already for you.

examples/virus_antibody/agents.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def move(self):
+        # If we've been removed from the space, bail out
+        if getattr(self, "space", None) is None:
```
quaquel, line 117: If implemented well, this should never be needed.

examples/virus_antibody/agents.py
```diff
@@ -0,0 +1,259 @@
...
+            )
+            self.direction = self.direction + perturb
+            norm = np.linalg.norm(self.direction)
+            if norm > 0:
+                self.direction /= norm
+            new_pos = self.position + self.direction * self.speed
+
+        # Chase a valid virus target
+        else:
+            if getattr(target, "space", None) is not None:
```
quaquel, line 151: again, this should not be needed if everything else is implemented correctly.

examples/virus_antibody/agents.py
```diff
@@ -0,0 +1,259 @@
...
+                    self.engage_virus(target)
+            else:
+                self.target = None
+
+        if new_pos is not None:
+            self.position = new_pos
+
+    def engage_virus(self, virus) -> str:
+        # If it's already gone
+        if virus not in self.model.agents:
```
quaquel, line 167: should not be needed.

examples/virus_antibody/agents.py
```diff
@@ -0,0 +1,259 @@
...
+            self.target = None
+            return "no_target"
```
quaquel, line 169: why return a string here?

examples/virus_antibody/agents.py
```diff
@@ -0,0 +1,259 @@
...
+            self.health -= 1
+            if self.health <= 0:
+                self.remove()
+                return "dead"
+
+            self.st_memory.append(dna)
+            self.lt_memory.append(dna)
+            self.ko_steps_left = self.ko_timeout
+            # mark KO state by weak-ref back to self
+            self.target = weakref.ref(self)
```
quaquel, line 187: I dont understand what is going on here.

examples/virus_antibody/agents.py
```diff
@@ -0,0 +1,259 @@
...
+        self.position = position
+        self.mutation_rate = mutation_rate
+        self.duplication_rate = duplication_rate
+        self.speed = 1
+        self.direction = np.array((1, 1), dtype=float)
+        self.dna = dna if dna is not None else self.generate_dna()
+
+    def step(self):
+        # If already removed from the space, don't do anything
+        if getattr(self, "space", None) is None:
```
quaquel, line 214: should not be needed.

examples/virus_antibody/agents.py
```diff
@@ -0,0 +1,259 @@
...
+                self.random.uniform(-0.5, 0.5),
+            ]
+        )
+        self.direction = self.direction + perturb
+        norm = np.linalg.norm(self.direction)
+        if norm > 0:
+            self.direction /= norm
+
+        # Step
+        self.position = self.position + self.direction * self.speed
```
quaquel, line 259:
```suggestion
        self.position += self.direction * self.speed
```

examples/virus_antibody/model.py
```diff
@@ -0,0 +1,153 @@
...
+            random=self.random,
+            # n_agents=initial_antibody + initial_viruses,
+        )
+
+        # Create and place the Antibody agents
+        antibodies_positions = self.rng.random(
+            size=(self.initial_antibody, 2)
+        ) * np.array(self.space.size)
+        directions = self.rng.uniform(-1, 1, size=(self.initial_antibody, 2))
+        self.antibodies_set = AntibodyAgent.create_agents(
```
quaquel, line 106: don't keep track of agents by type yourself.

examples/virus_antibody/model.py
```diff
@@ -0,0 +1,153 @@
...
+        )
+
+        # Create and place the Virus agents
+        dna = [self.random.randint(0, 9) for _ in range(3)]
+        viruses_positions = self.rng.random(size=(self.initial_viruses, 2)) * np.array(
+            self.space.size
+        )
+        directions = self.rng.uniform(-1, 1, size=(self.initial_viruses, 2))
+
+        self.viruses_set = VirusAgent.create_agents(
```
quaquel, line 125: same here, don't do this. It breaks the behavior of remove and is why you have all the if checks that are not needed.

quaquel (changes requested):
I have made a first pass. I like the idea of the model, but there is ample room for improvements. If you have questions, don't hesitate to ask.

examples/virus_antibody/agents.py
colinfrisch: I made the choice of limiting the parameter to these ones because otherwise it's a lot for the user in my opinion. If I add all of them, it would mean more than 10 parameters to be decided by the user. Do you think I should still add all of the other parameters ? Or are there one or two more that would be good to add and still leave the others hidden (like ko_timeout, sight_range, and a few non-crucial parameters) ?

colinfrisch:
Thanks a lot @quaquel for this first review ! I'm going to see what I can change asap

examples/virus_antibody/agents.py
quaquel: Fair enough. One option is to turn the constants that are the same for all agents of a given class into class level attributes. This still makes it easy to change them without having to add them all to the signature of the `__init__`.

quaquel:
> Thanks a lot @quaquel for this first review ! I'm going to see what I can change asap

I would start with the two sets in the the init of your model. Next, I would remove the if checks you have to see if an agent is still active in the model one at a time (or do one first to confirm the behavior is still correct). Again, don't hesitate to ask if you run into any issues.

examples/virus_antibody/agents.py
colinfrisch: Good idea ! It does allow to delete quite a few lines of code in `model.py` also.

examples/virus_antibody/agents.py
colinfrisch: I used it for debugging at first and it went past my checks before the PR, will be removed.

examples/virus_antibody/agents.py
colinfrisch: When an antibody looses a confrontation with a virus (because the virus DNA is not in it's memory yet), the antibody is ko for a few steps. It can't move during this time and I symbolise this by setting the agent's target to itself.

colinfrisch:
> I would start with the two sets in the the init of your model. Next, I would remove the if checks you have to see if an agent is still active in the model one at a time (or do one first to confirm the behavior is still correct). Again, don't hesitate to ask if you run into any issues.

I removed the checks and did the other fixes that you suggested. I still have one error that I don't understand, and it seems to be still linked to referencing. For certain configurations of parameters, the model runs for a while (even hundreds of steps) before making the error `error in step: 'NoneType' object has no attribute 'agent_positions'`. Even after debugging, I can't find the origin of it... would you have any idea to why it's happening ? 

You can reproduce this error by keeping the base parameters and setting `duplication rate` to max in the sliders.


Screenshot 2025-04-20 at 18 27 20


**Full traceback :** 

```Traceback (most recent call last):
  File "/Users/colinfrisch/Desktop/mesa/mesa/visualization/solara_viz.py", line 275, in step
    do_step()
    ~~~~~~~^^
  File "/Users/colinfrisch/Desktop/mesa/mesa/mesa_logging.py", line 132, in wrapper
    res = func(*args, **kwargs)
  File "/Users/colinfrisch/Desktop/mesa/mesa/visualization/solara_viz.py", line 309, in do_step
    model.value.step()
    ~~~~~~~~~~~~~~~~^^
  File "/Users/colinfrisch/Desktop/mesa/mesa/model.py", line 122, in _wrapped_step
    self._user_step(*args, **kwargs)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/Users/colinfrisch/Desktop/mesa-examples/examples/virus_antibody/model.py", line 131, in step
    self.agents.shuffle_do("step")
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/Users/colinfrisch/Desktop/mesa/mesa/agent.py", line 345, in shuffle_do
    getattr(agent, method)(*args, **kwargs)
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/Users/colinfrisch/Desktop/mesa-examples/examples/virus_antibody/agents.py", line 185, in step
    self.move()
    ~~~~~~~~~^^
  File "/Users/colinfrisch/Desktop/mesa-examples/examples/virus_antibody/agents.py", line 221, in move
    self.position += self.direction * self.speed
    ^^^^^^^^^^^^^
  File "/Users/colinfrisch/Desktop/mesa/mesa/experimental/continuous_space/continuous_space_agents.py", line 34, in position
    return self.space.agent_positions[self.space._agent_to_index[self]]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'agent_positions'
```

quaquel:
I'll try to take a closer look asap. It indeed seems that step is called on an agent that has already been removed.

examples/virus_antibody/agents.py
quaquel: Ok, clear. 

This might be implemented using event scheduling, which would result in a modest speedup of runtime.

examples/virus_antibody/agents.py
```diff
@@ -0,0 +1,225 @@
...
+        idx = self.random.randint(0, 2)
+        chance = self.random.random()
+        if chance < self.mutation_rate / 2:
+            dna[idx] = (dna[idx] + 1) % 10
+        elif chance < self.mutation_rate:
+            dna[idx] = (dna[idx] - 1) % 10
+        return dna
+
+    def move(self):
+        # Random walk
```
quaquel, line 213: Both agents have a random walk component. Might it be possible to implement this the same way for both and move it to a new super agent class from which both current agents derive?

colinfrisch: It could be a good idea ! At first I thought it could also be used for the `duplicate()` method. However, there are a lot of parameters to take into account and the code is very short inside, so I don't think that it would really help to make it derive from a parent class.

So I the end, there is just a few lines for the random movement. Isn't a whole parent class a bit overkill ?

quaquel: If the choice is between repeated code or a parent class, I prefer the parent class. Of course, in python you could also move the relevant code into a helper function (not really appropriate for random walks).

quaquel: Monday was a banking holiday so I had family obligations. I'll try to get to it today, but no promises unfortunately.

colinfrisch: Of course, nothing urgent ! Only if you have time :)

wang-boyu:
I'm not very familiar with how weakref works in AgentSet, so I could be wrong about this. If an agent's hard references are all removed, does its object guaranteed to be destroyed? If it's not, then there might be problems from here (and other similar places in `AgentSet.do()` and `AgentSet.shuffle_do()`):

https://github.com/projectmesa/mesa/blob/94c56a359b967da7b9d23e7df7bdc3ef446ab158/mesa/agent.py#L327

Suppose there are 50 agents, and Agent # 1 removed Agent # 7 in its `step()` function. Imagine after this, Agent # 7's object is not destroyed immediately, then `AgentSet.do()` or `AgentSet.shuffle_do()` will still try to run Agent # 7's `step()` function.

This may also happen if users unintentionally use hard references (e.g., `self.target = another_agent` instead of using a weakref that is shown in this PR). I understand https://github.com/projectmesa/mesa/discussions/2763 may fix this. However, there are just too many ways that hard references can happen, and I don't think it's possible to make sure that all references to Agent are weak refs. Examples include `self.some_dict[target] = another_agent`, `self.some_list.append(another_agent)`, or setting attributes in a non-Agent class.

Overall, I think we may need a better way to do null checks in `AgentSet.do()` and `AgentSet.shuffle_do()`.

In addition, there might be another problem in `AgentSet.do()` when we do

```python
for agentref in self._agents.keyrefs():
    ...
```

As agents are removed (deregistered), `keyrefs()` are updated. It is not really a good design to update iterable while looping through it. In contrast, `AgentSet.shuffle_do()` seems good by iterating a copy of `keyrefs()`:

```python
weakrefs = list(self._agents.keyrefs())
...
for ref in weakrefs:
    ...
```

quaquel:
> I'm not very familiar with how weakref works in AgentSet, so I could be wrong about this. If an agent's hard references are all removed, does its object guaranteed to be destroyed? 

This is indeed the central question. I just need time to dive into this and do some research. The random nature of the issue suggests that there is no hard guarantee that an agent is removed immediately upon its reference counter hitting 0. But I want to confirm this and read up on the garbage collection in Python (which is responsible for clearing up objects whose reference counter has hit 0).

EwoutH:
Awesome, if you want anything reviewed be sure to let me know!

EwoutH (approved):

PR #254, opened 2025-04-21
add meta-agent warehouse examples

tpike3:


tpike3:
I am going to violate protocol a little here and merge so the release notes for 3.2 work. 

Justification: 
- Ensure release note for 3.2 
- It is mesa-examples and not core mesa, although I hope to start updating mesa examples is in the next 2 months so it becomes more formal and standardized. (and the tests work)

PR #260, opened 2025-05-19
Sync the pre-commit config with Mesa

derkweijers:
Hi!

This PR syncs the pre-commit configuration from the main Mesa repo to the Mesa examples repo.
It adds the `.codespellignore` file as well, because pre-commit depends on it, and will fail (locally) if it's not present.

This PR links to issue [2782](https://github.com/projectmesa/mesa/issues/2782) but specifically targets the first item in the list in that issue, to keep the PRs small and organized.

### DIFF
--- a/.codespellignore
+++ b/.codespellignore
@@ -1,0 +1,10 @@
+hist
+hart
+mutch
+ist
+inactivate
+ue
+fpr
+falsy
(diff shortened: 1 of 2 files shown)


jackiekazil:
LGTM

PR #262, opened 2025-05-19
fix(examples): warehouse

### LINKED ISSUE #261: `warehouse` example won't run on command line
when i run
`solara run app.py`
i see
```

ERROR:    Application startup failed. Exiting.
```

dlovell:
fixes #261 

- move non-app files to `warehouse` so both `solara run app.py` and `pytest` work
- document `mesa>=3` requirement
- fix `color_patches` reference to `CellAgent` so that `pytest` can run


tpike3 (approved):
@dlovell Looks good to me, thanks!

PR #268, opened 2025-05-19
Sync pyproject with mesa

derkweijers:
This PR syncs parts of the `pyproject.toml` from the Mesa repo to this mesa-examples repo. It also updates some project/contact info.

The PR is a bit tricky since it introduces new Ruff rules. For example, upper-case variables are discouraged. Sometimes it makes sense though (for example if it's part of a formula). I've added ignore flags where applicable and fixed the other cases.

Furthermore, Ruff rule [B008](https://docs.astral.sh/ruff/rules/function-call-in-default-argument/) was triggered a couple of times. That's fixed in this PR as well, but I would appreciate opinions on how it's solved.

I've opted out of the Docstrings rule since that might be an entire PR on its own.


jackiekazil:
@derkweijers thank you for the contribution. Can you rebase and squash? And also fix the checks.

jackiekazil:
I spoke too soon. It looks like #271 fixes the ruff errors. 
Maybe that gets merged first?

tpike3 (approved):
Thanks @derkweijers!

PR #270, opened 2025-05-19
fix(examples): boltzmann wealth model network

jchanke:
updated the example to work with mesa 3.2.0, and the installation instructions.

mesa-boltzmann-network


tpike3 (approved):
LGTM! Thanks @jchanke!

Issue #263, opened 2025-05-19
Unnecessary placeholder directory


jacklinke:
# Background

In the `mesa-examples` repo, there is a non-existent subproject, "rl_models" (https://github.com/projectmesa/mesa-examples/blob/main/rl/rl_models) that returns a 404 Error page when navigating to it in the GitHub UI.

The directory was added in c4b4c75.

It seems to be a placeholder for the pre-weighted models hosted at (https://huggingface.co/projectmesa/rl_models/), which can be cloned into the current directory and used in the examples.

# Discussion

Because navigating to the folder results in a 404 error, it could cause confusion or decrease confidence in the Mesa examples for folks newly approaching the repo.

The only code relying on this directory to be present is `rl/Tutorials.ipynb`, which should be updated with instructions to perform the git clone of the actual `rl_models` repo prior to running the notebook.

# Recommendation

Remove the folder and update `rl/Tutorials.ipynb` and `rl/example.py` comments to reflect the actual location of the repo.

PR #275, opened 2025-05-20
Remove empty rl_models directory

ecourt93:
This fixes issue #263 .

I've removed the empty rl_models directory inside the rl folder. The README already seems to be up to date; it mentions that the hugging face model should be cloned. The expectation seems to be that users are working inside that rl directory, as intended, so no changes were made to the README.


jackiekazil:
Thank you!

PR #267, opened 2025-05-19
fix(examples): `bank_reserves`

dlovell:
update `bank_reserves` for `mesa>=3`

PR #276, opened 2025-05-20
fix(example): update readme for bank_transfer example

jchanke:
update the instructions to run the simulation (using Solara) and the URL to open it (using the updated port no.) 

small patch to @dlovell's fix for the `bank_reserves` example: #267.


tpike3 (approved):
LGTM

PR #298, opened 2026-01-07
Fix hotelling python 312

Tejasv-Singh:
@tpike3 I've updated the PR based on your feedback.

Dependencies: Reverted changes to requirements.txt and updated README.md to use pip install "mesa[rec]".

Logic: The pos property fixes (with setters) in agents.py remain in place to ensure agents render correctly on the grid.

Ready for review!


tpike3 (approved):
Thanks @Tejasv-Singh -- LGTM

PR #314, opened 2026-01-28
Fix for DeprecationWarning from `AgentSet.__getitem__`

quaquel:
Sequence behavior on AgentSet is being deprecated. So we call `self.agent.to_list()` before `random.sample` in `model.step`. This also gives a substantial performance boost for this model. 

Same fix in warehouse model, which still fails because sequence behavior is used in mesa.experimental's meta_agent stuff. The code in the example is now correct.

### DIFF
--- a/examples/deffuant_weisbuch/deffuant_weisbuch/model.py
+++ b/examples/deffuant_weisbuch/deffuant_weisbuch/model.py
@@ -60,7 +60,7 @@ class DeffuantWeisbuchModel(Model):
         """
         agent_list = self.agents.to_list()
         for _ in range(self.n):
-            agent_a, agent_b = self.random.sample(agents, 2)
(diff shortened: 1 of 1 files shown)


examples/warehouse/warehouse/model.py
```diff
@@ -95,8 +95,10 @@ def central_move(self, robot):
     def step(self):
         """Advance the model by one step."""
         for robot in self.agents_by_type[type(self.RobotAgent)]:
+            agents = self.agents_by_type[InventoryAgent].to_list()
+
             if robot.status == "open":  # Assign a task to the robot
-                item = self.random.choice(self.agents_by_type[InventoryAgent])
+                item = self.random.choice(agents)
```
EwoutH:
```suggestion
                item = self.random.choice(agent_list)
```

PR #317, opened 2026-01-30
fix: Add rng parameter to BankReservesModel

### LINKED ISSUE #316: BankReservesModel missing rng parameter causes batch_run to fail
**Describe the bug**
The `BankReservesModel` is missing the `rng` parameter in its `__init__()` method, which causes `batch_run.py` to fail with a `TypeError` when executed. The `mesa.batch_run()` function automatically injects the `rng` parameter for reproducibility, but the model doesn't accept it.

**Notably, bank_reserves is the only example in the repository that includes a `batch_run.py` file**, making this bug particularly important as it breaks the only demonstration of Mesa's batch_run functionality.

**Expected behavior**
Running `python batch_run.py` should complete successfully and generate a CSV file with data from all parameter combinations, similar to other Mesa examples that support batch runs.

**To Reproduce**
```bash
cd examples/bank_reserves
python batch_run.py
```
**Error Output** 
```
0%|                                                    | 0/4 [00:00<?, ?it/s]
  File "/path/to/mesa-examples/examples/bank_reserves/batch_run.py", line 49, in <module>
    main()
  File "/path/to/mesa-examples/examples/bank_reserves/batch_run.py", line 40, in main
    data = mesa.batch_run(
        BankReservesModel,
        br_params,
    )
  File "/path/to/mesa/mesa/batchrunner.py", line 122, in batch_run
    data = process_func(run)
  File "/path/to/mesa/mesa/batchrunner.py", line 204, in _model_run_func
    model = model_cls(**kwargs)
TypeError: BankReservesModel.__init__() got an unexpected keyword argument 'rng'
```
**Additional context**

- **Mesa version:** `>=3.1.4` (from requirements.txt)
- **Python version:** 3.10+ (tested on 3.13.5)
- **OS:** macOS/Linux/Windows (reproducible on all platforms)

**Root cause:**
This appears to be an incomplete part of the Mesa 3.x migration. Commit b23f89c (#267) updated the visualization from server.py to Solara but didn't add the `rng` parameter to the model's `__init__()` method.

**Pattern comparison:**
Other examples like `forest_fire`, `hotelling_law`, and `hex_snowflake` already follow the correct pattern by accepting `rng` or `seed` parameters and passing them to `super().__init__()`. 

**Why this wasn't caught:**
The bank_reserves example is the only one with a `batch_run.py` file in the examples directory, which is why this issue wasn't caught during the migration.

**Solution:**
I have a fix ready and will submit a PR shortly.

Jayantparashar10:
## Description
Fixes #316 

This PR adds the missing `rng` parameter to `BankReservesModel.__init__()` to make it compatible with `mesa.batch_run()`.

## Changes
- Added `rng=None` parameter to `BankReservesModel.__init__()`
- Pass `rng` to `super().__init__(rng=rng)` for proper RNG initialization
- Aligns with the pattern used in other examples (forest_fire, hotelling_law, hex_snowflake)

## Testing

- [x] Verified `batch_run.py` executes successfully without errors
- [x] Generates correct CSV output with all parameter combinations
- [x] Backward compatible with interactive visualization (`solara run app.py`)

## Context
This completes the Mesa 3.x migration for bank_reserves that was started in #267, which updated the visualization but missed the RNG parameter update. 

Since bank_reserves is currently the **only example with a batch_run.py file**, this fix is particularly important for demonstrating Mesa's batch_run functionality to users.

## Code Changes 

**Before** 
```python
def __init__(self, height=..., width=..., init_people=..., 
             rich_threshold=..., reserve_percent=...):
    super().__init__()  # Missing rng parameter
```
**After**
```python 
def __init__(self, height=..., width=..., init_people=..., 
             rich_threshold=..., reserve_percent=..., rng=None):
    super().__init__(rng=rng)   # Properly passes rng to parent
```

## Related
- Closes #316
- Related to #267 (Mesa 3.x migration)

### DIFF
--- a/examples/bank_reserves/bank_reserves/model.py
+++ b/examples/bank_reserves/bank_reserves/model.py
@@ -105,8 +105,9 @@ class BankReservesModel(mesa.Model):
         init_people=2,
         rich_threshold=10,
         reserve_percent=50,
+        rng=None,
     ):
-        super().__init__()
+        super().__init__(rng=rng)
         self.height = height
         self.width = width
         self.init_people = init_people


tpike3 (approved):
This is a fairly straightforward change and is consistent with #329

PR #320, opened 2026-02-03
refactor gis cell init to accept kwargs

wang-boyu:
This PR updates constructors of raster cell subclasses, from for example

```python
class LakeCell(mesa_geo.Cell):
    def __init__(self, model, pos=None, indices=None):
        super().__init__(model, pos, indices)
        ...
```

to

```python
class LakeCell(mesa_geo.Cell):
    def __init__(self, model=None, **kwargs):
        super().__init__(model, **kwargs)
        ...
```

This is mainly due to changes to raster cell constructor from https://github.com/mesa/mesa-geo/pull/299, where we deprecated `cell.indices` to `cell.rowcol`, and added `cell.xy`. In addition, when we initialize cells when creating a raster layer, this is how cell constructor is called now ([link to code](https://github.com/mesa/mesa-geo/blob/77200e8150301e5d96f0d2003750abf4151fcb74/mesa_geo/raster_layers.py#L338-L343)):

```python
cell = self.cell_cls(
    model,
    pos=(grid_x, grid_y),
    rowcol=(row_idx, col_idx),
    xy=xy,
)
```

where `cell_cls` is the user-defined subclass of raster cell, such as the `LakeCell` above.

By putting all parameters but `model` into `**kwargs`, this PR demonstrates how to subclass mesa-geo's raster cell class in case there are further changes to constructor API in the future.

### DIFF
--- a/gis/population/population/model.py
+++ b/gis/population/population/model.py
@@ -79,13 +79,13 @@ class Population(mesa.Model):
             if popu_round > 0:
                 for _ in range(popu_round):
                     num_agents += 1
-                    point = Point(self.space.population_layer.transform * cell.indices)
+                    point = Point(self.space.population_layer.transform * cell.rowcol)
                     if not point.within(self.space.lake):
                         person = Person(
                             model=self,
                             crs=self.space.crs,
                             geometry=point,
-                            img_coord=cell.indices,
+                            img_coord=cell.rowcol,
                         )
                         person.set_random_world_coord()
                         self.space.add_agents(person)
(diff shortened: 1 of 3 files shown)


EwoutH:
Thanks for the PR! I fixed another CI issue in 9567bdf, so CI should display the state of the example models again.

It seems to fix the main issue, however, this does expose a new issue:

```
DeprecationWarning: Cell.pos setter is deprecated and will be read-only in a future release.
```
Is that also related to this PR?

Further, can we do anything about this change being breaking to user models (see https://github.com/mesa/mesa-geo/pull/299#issuecomment-3890507417)?

EwoutH (approved):

PR #322, opened 2026-02-03
examples: add luck vs skill gambling model

riddhi2106:
### Summary
This PR adds a new example demonstrating how short-term gambling outcomes are dominated by luck, even when agents differ in skill. The example also shows that early winners are not reliably more skilled than early losers, illustrating why short-term success is often misattributed to ability.

### Motive
In stochastic environments, observed performance is often mistaken for underlying ability. This is especially common in gambling and similar settings, where short-term outcomes are highly noisy but are still used to infer skill.

This example is motivated by the need to clearly demonstrate why early success is an unreliable indicator of ability. By providing a minimal and transparent model, the example helps users understand how randomness and ranking alone can produce apparent performance differences, even when true skill differences are small.

The goal is to offer a simple, intuitive illustration of the distinction between luck and skill.

### Implementation
The example implements a minimal gambling model with the following structure:
- Each agent is assigned a fixed skill level drawn uniformly from [0,1].
- α (skill impact): A model parameter that controls how strongly an agent’s skill affects the probability of winning; small values mean outcomes are mostly driven by chance, while larger values make skill more influential.
- Skill slightly biases the probability of winning a bet, while outcomes remain stochastic.
- Agents repeatedly place fixed-size bets over a short time horizon.
- After each step, agents are ranked by wealth.
- The model tracks the average skill of the top 10% and bottom 10% of agents by wealth.

The visualization shows how, over short horizons, the average skill of winners and losers remains similar despite clear differences in outcomes.

### Usage Examples 
This is a screenshot of one of the possible outputs.
image

### DIFF
--- a/examples/gambling/model.py
+++ b/examples/gambling/model.py
@@ -39,8 +39,7 @@ class LuckVsSkillModel(Model):
     def step(self):
         self.steps += 1
 
-        for agent in list(self.agents):
-            agent.step()
+        self.agents.shuffle_do("step")
 
         self.datacollector.collect(self)
 


EwoutH (commented):
Thanks for your PR.

You placed all your files in the root of the repo, could you move it to it's own folder, like `examples/gambling/` ?

Issue #124, opened 2024-04-24
Use new Discrete Event Scheduling in examples


EwoutH:
_Open for contributions!_

Now that the Discrete Event Scheduler (DEVS) is released as experimental in Mesa 2.3.0, it would be interesting if we can use it in examples.

- Current code implementation: [mesa/experimental/devs](https://github.com/projectmesa/mesa/tree/main/mesa/experimental/devs)
- Original PR: https://github.com/projectmesa/mesa/pull/2066
- Original discussion: https://github.com/projectmesa/mesa/discussions/2032

All three mesa [benchmark models](https://github.com/projectmesa/mesa/blob/main/benchmarks) already use devs, so those changes could be ported over.

The approach for this would generally be:

- Read up on DEVS
- Go through the Mesa examples and seek opportunities where the DEVS could simplify or improve things.
- Open a Discussion on mesa-examples proposing your changes
- Implement them and open a PR here

The nuance here being that applying the DEVS is not a goal to be implemented as much as possible, but a means that could make some model (parts) more elegant, efficient, readable and/or performant.

EwoutH:
Bringing over and documenting the [`wolf_sheep` benchmark model](https://github.com/projectmesa/mesa/blob/main/benchmarks/WolfSheep/wolf_sheep.py) could be a good start.

sumana-2705:
hello @EwoutH 

I want to work on this issue, can you please assign it to me.

ShreyasNav:
Hi @EwoutH 
I’d like to work on this issue if it’s still available.
I have been studying mesa examples for a while and I think I can do this.
Please let me know if it’s okay for me to proceed

EwoutH:
@sumana-2705 @ShreyasNav sorry for the late reply, feel both free to work on this!

ygchavan1234:
@EwoutH I’ve been studying the performance bottlenecks in the WolfSheep model, specifically the grass polling overhead, and I'm interested in using this as a case study for my GSoC proposal.

I noticed that PR #3461 (Actions API) was just merged. Given the Mesa 4.0 roadmap, would the team prefer a refactor that implements grass regrowth as a timed Action within the new behavioral framework, or should I stick to a standard discrete-event approach?

I'm planning to document my benchmarks for both approaches in my GSoC-learning-space and would appreciate a quick nod on which direction provides more value to the Mesa 4.0 transition.

EwoutH:
Thanks for reaching out.

[...]

PR #323, opened 2026-02-04
feat: new "event based collisions" example

rico-haas:
### Summary
This is a playful example that makes use of Mesa's ~~DEVSimulator~~ event-based simulation capabilities on a continuous timeline. It features moving discs in a 2D environment that collide with the border and each other, using an event-based collision mechanism. With help of the simulator, discs will collide accurately, no matter how fast they are!

event_based_collisions

### Motive
This provides an example usage of Mesa's ~~DEVSimulator~~ event-based simulation capabilities on a continuous timeline, see #124. This model has been explicitly designed to demonstrate that ~~simulator~~ feature, since without it, the model couldn't have the same accuracy guarantees (discs could pass through each other).

### Implementation
1. Discs are placed in the space with a random movement direction.
2. At the beginning and whenever a disc changes its trajectory, the disc calculates when it could collide with an obstacle (other discs or the border) and schedules these as future events.
3. When such an event occurs and the disc still has the same trajectory (it hasn't collided with something else) then it changes its trajectory based on its angle of incidence.
4. Each full time step, all discs update their shown position based on their current trajectory. This doesn't update the actual model at all, it only queries the current state. Hence the simulation results are always the same, regardless of the timestep interval.

### Usage Examples
1. Set the disc speed to 1 and let the simulation run for let's say 100 time steps.
2. Take a screenshot of where all the discs are after 100 steps.
3. Turn the disc speed up to 10 and reset the simulation.
4. Now let the simulation run only 10 time steps.
5. Compare the result with your screenshot. It should be exactly the same!

### Additional Notes
This example is based on [my idea for a DEVS tutorial](https://github.com/mesa/mesa/issues/2196#issuecomment-3679076486).

### DIFF
--- a/examples/event_based_collisions/event_based_collisions/agents.py
+++ b/examples/event_based_collisions/event_based_collisions/agents.py
@@ -88,7 +88,7 @@ class DiscAgent(ContinuousSpaceAgent):
         Returns
         -------
         BorderCollisionEvent
-            To resolve te next border collision, this event's `resolve_collision`
+            To resolve the next border collision, this event's `resolve_collision`
             function should be called by the model at the time specified by the event's
             `collision_time` attribute.
         """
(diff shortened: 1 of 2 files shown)


EwoutH:
This seems really cool! I toyed around in the days I was still using NetLogo with almost exactly a model like this.

I'm a bit busy, but I will try to review somewhere this weekend. Maybe @tpike3 or @quaquel might also find this interesting.

PR #335, opened 2026-02-23
Remove unused deprecated ABMSimulator from virus_antibody example

ShyamAlancode:
While testing mesa-examples with the latest Mesa version, I noticed that
the virus_antibody example imports and instantiates ABMSimulator from
mesa.experimental.devs.

Mesa currently emits a FutureWarning indicating that ABMSimulator is
deprecated and will be removed in a future release.

In this example, the simulator object does not appear to be used
elsewhere in the file. I removed the unused import and instantiation.

After this change:

- The example launches normally
- The simulation runs as expected
- The ABMSimulator deprecation warning disappears

This is a small cleanup to help keep the example compatible with future
Mesa releases.

### DIFF
--- a/examples/virus_antibody/app.py
+++ b/examples/virus_antibody/app.py
@@ -1,7 +1,6 @@
 import weakref
 
 from matplotlib.markers import MarkerStyle
-from mesa.experimental.devs import ABMSimulator
 from mesa.visualization import (
     Slider,
     SolaraViz,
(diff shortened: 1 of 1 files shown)


ShyamAlancode:
> LGTM Thanks @ShyamAlancode 

Thanks for reviewing and merging, @tpike3 🙂 Glad the fix was helpful!

I’ll continue exploring the examples locally and let you know if I find anything else useful.

PR #340, opened 2026-02-27
Add M/M/c queue example as canonical DES model

EwoutH:
### Summary
Adds a classic M/M/c queuing model as the first pure discrete event simulation example, demonstrating Mesa's new event scheduling API without using `step()`.

### Motive
Mesa 3.5 introduced public event scheduling and time advancement (https://github.com/mesa/mesa/pull/3266), but we lack an example showing pure DES usage. The M/M/c queue is the "hello world" of discrete event simulation: simple, well-understood, and with analytical solutions for validation. It fills the gap between our ABM examples and the new DES capabilities.

### Implementation
- **`agents.py`**: `Customer` (created on arrival, removed after service) and `Server` (pulls from queue when idle — server-centric design).
- **`model.py`**: `MMcQueue` model using `schedule_recurring` with a callable interval for Poisson arrivals and `schedule_event(after=...)` for service completions. Disables the default step schedule via `_default_schedule.stop()`.
- **`analytical_mmc.py`**: Erlang C steady-state solutions for validation.

### Usage Examples
```python
from model import MMcQueue, analytical_mmc

model = MMcQueue(arrival_rate=2.0, service_rate=1.0, n_servers=3, rng=42)
model.run_until(10_000)

print(model.server_utilization)  # ≈ 0.667
print(model.avg_wait_time)       # ≈ 0.375

# Compare against analytical steady-state values
analytical = analytical_mmc(2.0, 1.0, 3)
```

Running `python model.py` prints a comparison table:
```
M/M/3 Queue (λ=2.0, μ=1.0, T=10000.0)
Customers served: 19992

Metric                     Simulated   Analytical
---------------------------------------------------
Server utilization              0.6672       0.6667
Avg wait time                   0.3716       0.3750
Avg system time                 1.3716       1.3750
```

### Additional Notes
- The model disables the default step schedule with `_default_schedule.stop()`. This works but may warrant a more explicit opt-out mechanism for pure DES models (e.g. a `Model` parameter).
- No visualization app is included yet, an implementation would be appreciated!

### DIFF
--- a/examples/mmc_queue/model.py
+++ b/examples/mmc_queue/model.py
@@ -33,6 +33,9 @@ class MMcQueue(Model):
     """
 
     def __init__(self, scenario=None, **kwargs):
+        if scenario is None:
+            scenario = MMcScenario(**kwargs)
+            kwargs = {}
         super().__init__(scenario=scenario, **kwargs)
 
         # Queue


examples/mmc_queue/model.py
```diff
@@ -0,0 +1,132 @@
...
+        self._default_schedule.stop()
+
+        # Schedule stochastic arrivals
+        self.schedule_recurring(
+            self._customer_arrival,
+            Schedule(
+                interval=lambda m: m.rng.exponential(1.0 / m.arrival_rate),
+                start=0.0,
+            ),
+        )
```
quaquel: what about adding the new data collection to this example. Your use of properties should make this quite easy.

quaquel:
> Implemented the Scenario. Had to use a weird workaround (https://github.com/mesa/mesa-examples/commit/9a9fbc1f5a1171a8430c9322ff7594538d258d7e). We might want to address that.

See https://github.com/mesa/mesa/pull/3493 for my fix for that.

PR #348, opened 2026-03-02
Fix conways_game_of_life_fast: resolve visualization crash with no agents

### LINKED ISSUE #285: conways way of life is broken?
attempted to run conways game of life

Image

Harshini2411:
Fixes #285

## Problem
Running app.py crashed with:
IndexError: too many indices for array: array is 1-dimensional, but 2 were indexed

Root cause: Mesa 3.3.1's _scatter() crashes when there are no agents 
because loc[:, 0] fails on an empty 1D array.

### model.py
- Added OrthogonalMooreGrid as self.grid (required for visualization)
- Changed standalone PropertyLayer to grid.create_property_layer()
- Added rng=None parameter for Mesa compatibility

### app.py
- Replaced broken make_space_component with a custom Solara matplotlib 
  component that renders cell_layer.data directly using imshow
- This bypasses Mesa 3.3.1's _scatter() crash on empty agent sets

## Tested
- Simulation runs and animates correctly
- Sliders (width, height, alive_fraction) work
- Reset works correctly

### DIFF
--- a/examples/conways_game_of_life_fast/app.py
+++ b/examples/conways_game_of_life_fast/app.py
@@ -18,13 +18,35 @@ def make_grid_component():
         plt.tight_layout()
         solara.FigureMatplotlib(fig)
         plt.close(fig)
+
     return GridComponent
 
 
 model_params = {
-    "width": {"type": "SliderInt", "value": 30, "label": "Width", "min": 5, "max": 60, "step": 1},
-    "height": {"type": "SliderInt", "value": 30, "label": "Height", "min": 5, "max": 60, "step": 1},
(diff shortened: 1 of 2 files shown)


Harshini2411:
@EwoutH Would appreciate a review when you get a chance! 🙂

quaquel:
This example has been updated to reflect changes coming in mesa 4. These are backward incompatible with mesa 3.x. 
Also, it now fails for both mesa 3.4 and mesa 4.

Harshini2411:
@quaquel Since it still fails on Mesa 4, would a fix be welcome? What approach would you recommend?

Harshini2411:
Thanks for clarifying @quaquel! I understand now, I'll close this PR since it was breaking working code rather than fixing anything. I'll be more careful to test against the correct Mesa version before raising a PR next time. Appreciate the patience!

PR #359, opened 2026-03-05
Adds a fast-spreading pathogen model with quarantine and compliance

### LINKED ISSUE #356: Add: Fast-Spreading Pathogen Model with Quarantine Compliance Mechanics
## Summary
A SIR-based pathogen outbreak model with no exposed/latency period — 
inspired loosely by Plague Inc, though the goal here is containment 
rather than world domination. The model simulates how quarantine(lockdown) 
compliance rates at the population level affect outbreak dynamics, analyzing 
changes in infection, death and immunity rates.

## Motivation
Started building this after the Boltzmann tutorial felt too simple and 
I wanted to model something with real behavioural stakes. It was inspired 
from an old game called Plague Inc, although the idea is completely different — 
thats where it began to take its shape. Mesa-examples currently has 
nothing that shows:

- Individual agentic behavioural response to disease outbreak (flee vs ignore quarantine)
- A two-threshold system for quarantine analysis in ABM
- Agent flee mechanics based on distance maximisation during quarantine

The existing `virus_antibody` model works at the cellular/biological level. 
This one sits at the population level - citizens making movement decisions 
in response to a spreading outbreak, driven by behaviour rather than 
policy intervention.

## What This Model Demonstrates

- **Two-threshold quarantine system** - quarantine triggers when infected 
  count exceeds an upper threshold and lifts only when it drops below a 
  lower one, this prevents quarantine to be removed the moment someone recovers which is 
  not realistic and which occurs with a single 
  threshold variable.

- **Compliance rate** - a configurable fraction of citizens actually follow 
  quarantine orders, modelling real-world partial adherence. The rest 
  (shown as squares) keep moving freely against orders.

- **Flee behaviour** - compliant healthy agents use Manhattan distance 
  maximisation to move away from infected agents during quarantine.

- **Infected agents freeze** - simulating a government isolating an infected 
  zone. Non-compliant agents ignore this entirely.

- **Immunity/recovery** - agents recover to full immunity after 10 steps(for now) 
  of being infected with a 1% death probability. Dead agents remain 
  on the grid as a visual indicator of outbreak severity.

- **Outcomes** - compliance rate is the main variable it determines at what level 
  the quarantine actually contain the outbreak also it can be to analyze how does changing 
  population size affect infection and death rates as healthy agents 
  have more room to flee from quarantined zones.

## Research Questions

- At what quarantine compliance rate does behavioural policy step in to intervene 
  meaningfully to slow down a fast-spreading pathogen?
- How does population density interact with flee behaviour i.e do agents 
  have enough room to actually escape?
- What happens when the quarantine threshold is set too high and triggers 
  too late or the lower threshold is too low for the virus starts to spread again?

## Mesa Features Used
- `MultiGrid`
- `DataCollector`
- `SolaraViz` with `make_space_component` and `make_plot_component`
- `AgentSet.do()`
- Custom Solara components for live quarantine status indicator

## Mesa Version
3.5

## PR
Will follow shortly.

Syn-Eon:
## Pathogen Outbreak and Quarantine Compliance Model

Adds a fast-spreading pathogen model with automatic 
quarantine and with it compliance rate mechanics.

Resolves #356

### Mesa features used
- MultiGrid, DataCollector, SolaraViz
- make_space_component, make_plot_component
- graphical mapping per step with a quarantine indicator
- with the latest mesa 3.5 code structure 

### Mesa version
4

### How to run
pip install -r requirements.txt
solara run app.py

@jackiekazil @EwoutH

## Demo of no compliance model 
no_compliance


EwoutH:
Thanks for the PR, looks like an interesting model.

Could you:
- Fix the test failure (a rebase might be enough)

abhinavk0220:
**Peer Review — Pathogen Outbreak & Quarantine Compliance Model**

Reviewed `agents.py`, `model.py`, `app.py`, and `README.md` in full.

---

**Overall**

The two-threshold quarantine hysteresis is a genuinely interesting mechanic  most epidemic examples use a single threshold and get oscillating quarantine on/off behaviour. The compliance + flee behaviour is well thought out and the four-screenshot README is one of the clearest I've seen for demonstrating parameter sensitivity. A few correctness issues before merge.

---

**Critical: Non-seeded RNG throughout `agents.py`**

```python
import random
...
self.compliant = random.random() < self.model.compliance_rate
if random.random() > 0.40:
    self.state = "infected"
chance_of_death = random.random()
```

Every stochastic decision uses Python's global `random` module instead of `self.random` (Mesa's seeded RNG). This makes the model completely non-reproducible — running with `rng=42` twice will give different results. Replace all `random.random()` calls with `self.random.random()`.

---

**Missing package structure**

`agents.py` and `model.py` live directly in `examples/pathogen_outbreak/` with no subdirectory package. All other examples in this repo use a pattern like `examples/my_model/my_model/agents.py`. `app.py` then does `from model import PathogenModel` (a relative import that only works when `app.py` is run from its own directory). This should be `from pathogen_outbreak.model import PathogenModel` with a proper `pathogen_outbreak/__init__.py`.

---

**`DataCollector.collect` not called at `__init__`**

`model.py` calls `self.datacollector.collect(self)` as the first thing in `step()`, but never in `__init__`. This means the time-series chart starts at step 1 with no step-0 baseline. Adding `self.datacollector.collect(self)` at the end of `__init__` is standard Mesa practice and gives a clean starting point.

---

**O(n) infected-position scan inside every compliant agent's `quarantine()`**

```python
for i in self.model.agents:
    if i.state == "infected":
        infected_pos.append(i.cell.coordinate)
```

Every compliant healthy agent scans all agents to find infected positions each step. With 300 agents during quarantine, this is O(n²) per step. A simple fix: cache `infected_positions` as a model-level attribute computed once at the start of `step()`, then reuse it in each agent call.

---

**README says MultiGrid, code uses `OrthogonalMooreGrid`**

> *citizens are placed randomly on a MultiGrid*

The model uses `OrthogonalMooreGrid` (Mesa 4.x API), not `MultiGrid` (removed in Mesa 4.x). Update the README to match the actual implementation.

---

**Dead agents block movement**

Dead agents remain on the grid (intentional, per README) but since `is_empty` returns `False` for cells they occupy, they silently block other agents from moving to those cells. This affects the flee and random-move logic. It's worth noting this in the README so the user understands why high-death scenarios can produce movement deadlocks.

---

**`quarantine` chart in `graph_ot`**

```python
graph_ot = make_plot_component(["healthy", "infected", "immune", "dead", "quarantine"])
```

`quarantine` is a model reporter (returns 0 or 1), not an agent reporter. `make_plot_component` with a list uses model reporters, so this is fine  but it would be clearer to explicitly pass a dict with labels and colors rather than a plain list.

---

**Summary**

The model concept is strong and the README is excellent. The two things that need fixing before merge: (1) replace all `random.random()` with `self.random.random()` for reproducibility, and (2) restructure into a proper package directory. Everything else is advisory.

Syn-Eon:
> **Peer Review — Pathogen Outbreak & Quarantine Compliance Model**
> 
> Reviewed `agents.py`, `model.py`, `app.py`, and `README.md` in full.
> 
> **Overall**
> 
> The two-threshold quarantine hysteresis is a genuinely interesting mechanic most epidemic examples use a single threshold and get oscillating quarantine on/off behaviour. The compliance + flee behaviour is well thought out and the four-screenshot README is one of the clearest I've seen for demonstrating parameter sensitivity. A few correctness issues before merge.
> 
> **Critical: Non-seeded RNG throughout `agents.py`**
> 
> ```python
> import random
> ...
> self.compliant = random.random() < self.model.compliance_rate
> if random.random() > 0.40:
>     self.state = "infected"
> chance_of_death = random.random()
> ```
> 
> Every stochastic decision uses Python's global `random` module instead of `self.random` (Mesa's seeded RNG). This makes the model completely non-reproducible — running with `rng=42` twice will give different results. Replace all `random.random()` calls with `self.random.random()`.
> 
> **Missing package structure**
> 
> `agents.py` and `model.py` live directly in `examples/pathogen_outbreak/` with no subdirectory package. All other examples in this repo use a pattern like `examples/my_model/my_model/agents.py`. `app.py` then does `from model import PathogenModel` (a relative import that only works when `app.py` is run from its own directory). This should be `from pathogen_outbreak.model import PathogenModel` with a proper `pathogen_outbreak/__init__.py`.
> 
> **`DataCollector.collect` not called at `__init__`**
> 
> `model.py` calls `self.datacollector.collect(self)` as the first thing in `step()`, but never in `__init__`. This means the time-series chart starts at step 1 with no step-0 baseline. Adding `self.datacollector.collect(self)` at the end of `__init__` is standard Mesa practice and gives a clean starting point.
> 
> **O(n) infected-position scan inside every compliant agent's `quarantine()`**
> 
> ```python
> for i in self.model.agents:
>     if i.state == "infected":
>         infected_pos.append(i.cell.coordinate)
> ```
> 
> Every compliant healthy agent scans all agents to find infected positions each step. With 300 agents during quarantine, this is O(n²) per step. A simple fix: cache `infected_positions` as a model-level attribute computed once at the start of `step()`, then reuse it in each agent call.
> 
> **README says MultiGrid, code uses `OrthogonalMooreGrid`**
> 
> > _citizens are placed randomly on a MultiGrid_
> 
> The model uses `OrthogonalMooreGrid` (Mesa 4.x API), not `MultiGrid` (removed in Mesa 4.x). Update the README to match the actual implementation.
> 
> **Dead agents block movement**
> 
> Dead agents remain on the grid (intentional, per README) but since `is_empty` returns `False` for cells they occupy, they silently block other agents from moving to those cells. This affects the flee and random-move logic. It's worth noting this in the README so the user understands why high-death scenarios can produce movement deadlocks.
> 
> **`quarantine` chart in `graph_ot`**
> 
> ```python
> graph_ot = make_plot_component(["healthy", "infected", "immune", "dead", "quarantine"])
> ```
> 
> `quarantine` is a model reporter (returns 0 or 1), not an agent reporter. `make_plot_component` with a list uses model reporters, so this is fine but it would be clearer to explicitly pass a dict with labels and colors rather than a plain list.
> 
> **Summary**
> 
> The model concept is strong and the README is excellent. The two things that need fixing before merge: (1) replace all `random.random()` with `self.random.random()` for reproducibility, and (2) restructure into a proper package directory. Everything else is advisory.

Thanks for the  review, really appreciate it. I fixed the two critical issues i.e I restructured the files into a proper package directory and used mesas own random function for proper seeding. Also added a note about dead agents blocking movement to the README file. Once again thanks for the review

catherinedevlin:
Made some small changes at https://github.com/mesa/mesa-examples/pull/468 to put this over the finish line.

PR #367, opened 2026-03-08
Add Axelrod Culture Model example

Punya-B-R:
### Summary
Adds an implementation of Axelrod's Culture Model, a classic ABM where agents on a grid hold multi-feature cultural profiles and interact with neighbors based on cultural similarity, producing emergent cultural regions.

### Motive
Mesa examples currently focus on spatial and epidemic models. This example demonstrates a culturally-driven diffusion process with continuous interaction probabilities, adding diversity to the example set.

### Implementation
- agents.py: CultureAgent holds F cultural features each with Q possible traits. Agents interact with a random neighbor with probability equal to their cultural similarity, copying one differing trait if interaction occurs.
- model.py: OrthogonalVonNeumannGrid setup, random culture initialization, cultural region tracking via BFS flood fill.
- app.py: Solara visualization with a color-coded culture grid and region count over time.

### Usage Examples
Run with: solara run app.py

Key parameters:
- F (features): higher F leads to faster convergence
- Q (traits per feature): higher Q leads to more fragmentation

### Additional Notes
Reference: Axelrod, R. (1997). The dissemination of culture: A model with local convergence and global polarization. Journal of Conflict Resolution, 41(2), 203-226.


EwoutH:

Could you:
- Checkout the test failure
- Add a screenshot or gif of the visualisation

abhinavk0220:
**Peer Review — Axelrod Culture Model**

Reviewed `agents.py`, `model.py`, `app.py`, and `README.md` in full.

---

**Overall**

A clean, well-scoped implementation of a classic ABM. The code is short, readable, and the Monte Carlo activation (N random interactions per step) correctly mirrors Axelrod's original specification. A few issues worth addressing before merge.

---

**Floating-point equality comparison in `interact_with`**

```python
if sim == 0.0 or sim == 1.0:
    return
```

`sim` is computed as `matches / len(self.culture)`  a rational number over integers, so it will always be exactly 0.0 or 1.0 with no floating-point rounding. This is safe here. However the `sim == 1.0` branch is redundant: if `sim == 1.0` there are no differing features, so `differing` would be an empty list and `self.random.choice(differing)` would raise an `IndexError`. The guard is necessary but a comment explaining this would prevent future confusion. Alternatively, checking `if not differing: return` after computing it is more explicit.

---

**`create_agents` call signature**

```python
CultureAgent.create_agents(self, width * height, cultures)
```

In Mesa 4.x, `Agent.create_agents(model, n, **kwargs)` distributes keyword arguments. `cultures` is being passed as a positional argument here (third position). This will likely fail or silently misbehave depending on Mesa's internal `create_agents` signature. The correct call should be `CultureAgent.create_agents(self, width * height, culture=cultures)` if Mesa supports per-agent list distribution, otherwise agents need to be created in a loop assigning cultures individually. Worth verifying this runs without error.

---

**Von Neumann neighborhood is correct**

Using `OrthogonalVonNeumannGrid` matches Axelrod's original model which uses 4-connected neighbors. ✅

---

**Color mapping via SHA256**

```python
h = hashlib.sha256(key).hexdigest()
return (int(h[0:2], 16) / 255, ...)
```

Clever approach, but SHA256 distributes uniformly across the hash space  culturally similar agents (differing in only one trait) will have completely unrelated colors, making it hard to visually track convergence. A perceptually-ordered mapping (e.g., indexing into a colormap by cultural distance from a reference) would make the visualization more informative. Not a blocker, just worth considering.

---

**Missing newline at end of `README.md`**

Both `README.md` and the repo-level README update end without a trailing newline (`\ No newline at end of file`). Minor but causes a git diff warning and can confuse some tools.

---

**DataCollector collects at `__init__` ✅**

Calling `self.datacollector.collect(self)` at the end of `__init__` is correct Mesa practice — gives step-0 data. ✅

---

**`rng` as `InputText` in model params**

Same note as other examples: exposing `rng` as a free-text field risks silent failure if a non-integer is entered. A `SliderInt` would be safer.

---

**Summary**

The model logic faithfully implements Axelrod (1997) and the code is clean. The main thing to verify before merge is whether `create_agents` handles the `cultures` list correctly as a positional argument if it doesn't, agents will all get the same (full list) culture which would break the model silently. Once that's confirmed or fixed, this is close to merge-ready.

Punya-B-R:
@EwoutH Thanks for the feedback! I fixed the test failure by switching to CellAgent with proper cell assignment. GIF attached below. I've also addressed the peer review points.
Axelrod Culture Model

Punya-B-R:
@abhinavk0220 Thanks for the review! Addressed most points:

- Switched to CellAgent for proper cell placement (this fixed the create_agents issue too)
- Added comment explaining the sim == 1.0 guard
- Changed rng to SliderInt
- Fixed README trailing newline
- Kept the SHA256 color mapping for now, happy to improve it in a follow-up

abhinavk0220:
Thanks for addressing all the points so quickly @Punya-B-R!

Switching to `CellAgent` was the right call — it resolves the `create_agents` signature concern cleanly and is the correct Mesa 4.x pattern. The GIF does a great job showing the cultural clustering emerge over time.

Keeping the SHA256 color mapping for now is fine — the model works correctly, that's a visual enhancement for a follow-up.

From my end the review is complete. The model logic is faithful to Axelrod (1997), Mesa APIs are used correctly, and the self-review checklist is thorough. Good work getting this to a solid state.

Punya-B-R:
@abhinavk0220 Thanks for the thorough review and for following up, really helpful feedback throughout!

jackiekazil:
This was merged in another PR see #466

PR #369, opened 2026-03-09
Add Smart Traffic Light Example

FedericoCelauro:
### Summary
This PR adds a new example model.  It's a very simple modellization of traffic flow at an intersection that can show how smart traffic light control can reduce the wait time of the cars.

### Motive
This is a very easy traffic model that can be made using meta agents.

### Implementation

In the model, both cars and traffic lights are agents that interact with each others. On top of them,  I implemented a meta-agent called IntersectionController, that can change the status of the traffic lights based on the  number of cars waiting in line.

### Usage Examples

To run the example usage do

```
    $ cd examples/smart_traffic_lights
    $ python run_example.py
```

To run the interactive Solara simulation please do

```
    $ cd examples/smart_traffic_lights
    $ solara run app.py
```

### Additional Notes
Being a simple model, this can be made much more complex implementing, for example, smarter traffic lights controls, larger road systems, non periodic boundary conditions and many more. It can be easily implemented using experimental meta agent class

### DIFF
--- a/examples/smart_traffic_lights/smart_traffic_lights/model.py
+++ b/examples/smart_traffic_lights/smart_traffic_lights/model.py
@@ -72,13 +72,15 @@ class TrafficModel(mesa.Model):
         for _ in range(num_cars_east):
             car = CarAgent(self, Direction.EAST)
             pos = (self.random.randrange(width), center_y)
-            car.cell = self.cells[pos]
+            car_cell = self.cells[pos]
+            car.move_to(car_cell)
 
         # Spawn North-bound cars on the vertical road
         for _ in range(num_cars_north):
             car = CarAgent(self, Direction.NORTH)
             pos = (center_x, self.random.randrange(height))
-            car.cell = self.cells[pos]
+            car_cell = self.cells[pos]
+            car.move_to(car_cell)
 
         # Setup Data Collection
         self.datacollector = DataCollector(


EwoutH:

Could you:
- Checkout the test failure
- Add a screenshot or gif of the visualisation to the PR description

FedericoCelauro:
I have fixed the version problems. 

Here is the Solara visualization of the model's behaviour.

chrome-capture-2026-03-09_TrafficModel_Mesa

examples/smart_traffic_lights/smart_traffic_lights/agents.py
```diff
@@ -0,0 +1,168 @@
...
+            elif isinstance(obj, TrafficLightAgent):
+                # Only stop if the light controls our direction and is red
+                if obj.direction == self.direction and obj.state == LightState.RED:
+                    can_move = False
+                    stopped_by_red_light = True
+                    break
+
+        if can_move:
+            # Moving the agent is now just reassigning the cell!
+            self.cell = next_cell
```
Syn-Eon: in the step function self.cell = next_cell is being used here to reassign or move the cars to the next cell, instead doing so, you can try using self.move_to(next_cell), seems more apt, or say the mesa way of moving as what is done here bypasses mesas tracking system.

examples/smart_traffic_lights/smart_traffic_lights/model.py
```diff
@@ -0,0 +1,115 @@
...
+        for _ in range(num_cars_east):
+            car = CarAgent(self, Direction.EAST)
+            pos = (self.random.randrange(width), center_y)
+            car.cell = self.cells[pos]
+
+        # Spawn North-bound cars on the vertical road
+        for _ in range(num_cars_north):
+            car = CarAgent(self, Direction.NORTH)
+            pos = (center_x, self.random.randrange(height))
+            car.cell = self.cells[pos]
```
Syn-Eon: same as line 82 of agents.py you could use move_to() function instead of reassigning

examples/smart_traffic_lights/smart_traffic_lights/model.py
```diff
@@ -0,0 +1,115 @@
...
+            self, smart=smart_lights, lights=[light_east, light_north], sensor_range=6
+        )
+
+        # Spawn Cars
+
+        # Spawn East-bound cars on the horizontal road
+        for _ in range(num_cars_east):
+            car = CarAgent(self, Direction.EAST)
+            pos = (self.random.randrange(width), center_y)
+            car.cell = self.cells[pos]
```
Syn-Eon: same as line 82 of agents.py you could use move_to() function instead of reassigning

examples/smart_traffic_lights/smart_traffic_lights/agents.py
```diff
@@ -0,0 +1,168 @@
+import enum
+
+import mesa
+from mesa.discrete_space import CellAgent
+
+
+class Direction(enum.Enum):
```
Syn-Eon, line 7: the Direction enum class only defines east and north, there's no south or west. This means traffic only flows in two directions. Is this intentional? A real intersection would have opposing traffic flows. This is just a suggestion from my side, not something to change, i think it would make the model more interesting whilst adding very less code.

examples/smart_traffic_lights/smart_traffic_lights/agents.py
```diff
@@ -0,0 +1,168 @@
...
+        if can_move:
+            # Moving the agent is now just reassigning the cell!
+            self.cell = next_cell
+        else:
+            self.total_wait_time += 1
+            if stopped_by_red_light:
+                self.red_light_wait_time += 1
+
+
+class IntersectionController(mesa.Agent):
```
Syn-Eon, line 89: extending mesa.Agent seems intentional since the controller has no physical grid position, but a comment explaining this design choice would be extremely helpful in understanding why it differs from the other agents specifically CellAgent.

examples/assets/chrome-capture-2026-03-09_TrafficModel_Mesa.gif
Syn-Eon, line 1: Found the model really interesting/unique had some fun playing around with it, also the addition of a Gif rather than a screenshot is really helpful.

FedericoCelauro:
Thank you @Syn-Eon for your comments! I'll certainly use them to improve my work. I'll check out your model as soon as I can.

examples/smart_traffic_lights/smart_traffic_lights/agents.py
FedericoCelauro: Yes, for simplicity I only had these two directions. Something that I want to implement is a grid-like street system with cars that can also change direction. This will get a more realistic model, but it requires some more work to make.

PR #370, opened 2026-03-09
pyproject: sync with mesa/mesa and ensure tests run against GitHub repo

annapurna-gupta:
Closes https://github.com/mesa/mesa/issues/2782 (pyproject sync checklist item)

Changes:
- Add `allow-direct-references = true` for hatchling
- Sync ruff target-version to py312
- Add codespell section
- Add extend-exclude = ["build"] to ruff
- Install mesa from GitHub main in all test dependency groups

### DIFF
--- a/gis/agents_and_networks/src/space/campus.py
+++ b/gis/agents_and_networks/src/space/campus.py
@@ -1,6 +1,5 @@
 import random
 from collections import defaultdict
-from typing import DefaultDict
 
 import mesa_geo as mg
 from shapely.geometry import Point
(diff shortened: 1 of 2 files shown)


annapurna-gupta:
> Thanks for the PR. Could you explain all changes in the PR description?
> 
> I think most of this makes sense. @quaquel, do you see anything weird?

Thanks for the review!

1. Added allow-direct-references = true so that dependencies specified via Git URLs can be used.
2. Sync Ruff target-version to py312 — synced from mesa/mesa main so that linting rules stay consistent.
3. Add ignore-words-list for codespell — synced from mesa/mesa main to ensure consistent spell-check behavior.
4. Add extend-exclude = ["build"] — tells Ruff to skip the auto-generated build/ folder so it only checks actual source code.
5. Install Mesa from GitHub main in all test dependency groups — this ensures tests run against the latest Mesa development version rather than a potentially outdated PyPI release.

PR #388, opened 2026-03-15
Add : bdi_recommender model

champ-byte:
This is a mesa implementation of BDI Recommender Agent from [NetLogo Modeling Commons](https://modelingcommons.org/browse/one_model/4602).

It is a good minimal example of Belief Desire Intention(BDI) architecture

This model simulates an agent ("Bob") who autonomously reasons about his health by maintaining **Beliefs** (knowledge about his health), **Desires** (motivations), **Goals** (achievable desires), and **Intentions** (action plans). It demonstrates how new information (advice from a Doctor) can dynamically change Bob's beliefs and alter his behavior.

Visualization:
image
image

###  Review Checklist
Does it belong?
 

- [x] No significant overlap with existing examples
- [x]  Well-scoped simplest model that demonstrates the idea
- [x]  Showcases Mesa features not already well-covered
- [x]  Showcases interesting ABM mechanics (LLM reasoning vs rule-based dynamics)

Is it correct?
 

- [x] Uses current Mesa APIs (OrthogonalMooreGrid, DataCollector)
- [x]  Runs and visualizes out of the box (does not require api key)
- [x]  Deterministic with fixed rng seed LLM outputs are non-deterministic by nature
- [x]  Moved to examples directory

Is it clean?

- [x]  No dead code or unused imports
- [x]  Clear naming, logic readable
- [x]  README explains what it does, what it demonstrates, how to run it
- [x]  PR follows template, commits reasonably clean


EwoutH:
Thanks for the PR, nice to see some BDI.

Could you:
- Add a screenshot or gif of the visualisation to the PR description
- Request a [peer-review](https://github.com/mesa/mesa-examples/issues/390) (and maybe do one or two yourself)

examples/bdi_recommender/model.py
```diff
@@ -0,0 +1,120 @@
...
+        # Establish relationship between user and doctor
+        self.user.doctor = self.doctor
+
+        # Data collection
+        self.datacollector = DataCollector(
+            agent_reporters={
+                "beliefs": lambda a: dict(a.beliefs) if hasattr(a, "beliefs") else {},
+                "trust": lambda a: a.trust if hasattr(a, "trust") else 0,
+                "goals": lambda a: (
+                    dict(a.goals) if hasattr(a, "goals") and callable(a.goals) else {}
```
ShreyasN707: the DataCollector incorrectly checks if `goals` is callable.  
Since `goals` is a `@computed_property`, it already returns a value, so `callable()` is always false.

examples/bdi_recommender/app.py
```diff
@@ -0,0 +1,139 @@
...
+    },
+}
+
+model = BDIRecommenderModel(rng=42)
+
+renderer = SpaceRenderer(model, backend="matplotlib").setup_agents(agent_portrayal)
+renderer.draw_agents()
+
+page = SolaraViz(
+    model,
```
ShreyasN707, line 133: Passing a model instance here might limit reactivity—using the model class could help updates work better.

ShreyasN707 (commented):
I have tested it in my local system and it works good with all the different model parameters i tired .

examples/bdi_recommender/model.py
champ-byte: Hi , thanks for looking into this , I will update it

examples/bdi_recommender/app.py
champ-byte: I think this is not an issue because , SolaraViz accepts `model: Model | solara.Reactive[Model]`

ShreyasN707: yes `SolaraViz` accepts both. My concern was more about reactivity in practice (e.g. model resets/param updates), but if it's working correctly with the current setup then it should be fine.

abhinavk0220:
**Peer Review — BDI Recommender Agent**

Reviewed `agents.py`, `model.py`, `app.py`, and `Readme.md` in full.

---

**Overall**

A genuinely interesting example BDI architecture is underrepresented in Mesa examples and the NetLogo port is a good choice. The use of `mesa_signals` (`Observable`, `computed_property`) to make beliefs reactive is the right architectural decision and showcases a Mesa feature that barely any other example uses. A few correctness issues need addressing before merge.

---

**Critical: Non-seeded RNG in `agents.py`**

`self.compliant = random.random() < self.model.compliance_rate` uses Python's stdlib `random` module directly. This breaks reproducibility running with `rng=42` will produce different compliance assignments on each run. Replace with `self.model.random.random()` (or pass `self.random` once `super().__init__()` is called). Same issue appears in the infection check `if random.random() > 0.40`.

---

**Trust Logic in `evaluate_proposal`**

```python
if sender and sender.trust > 0.5:
```

This checks the *doctor's* trust level, not Bob's trust in the doctor. The BDI model says Bob's receptiveness should be governed by *his own* trust variable (already stored on `self.trust`). The current code means the proposal is accepted based on whether the doctor trusts himself, which is always true since the doctor's trust initialises above 0.5. This should likely be `self.trust > 0.5`.

---

**`goals` computed property recomputes on every access**

`elect_goals()` iterates through all desires and beliefs on every access. In `step()`, `self.goals` is called in both `step()` and `update_beliefs()` so it runs at least twice per step. This is fine for a small model but worth noting; caching the result until beliefs change would be cleaner given you already have `Observable` beliefs.

---

**Hardcoded location coordinates in `model.py`**

```python
self.gym = GymAgent(self, self.grid[21, 26])
self.nutrition_centre = NutritionCentreAgent(self, self.grid[24, 22])
self.clinic = ClinicAgent(self, self.grid[16, 19])
```

These are hardcoded for a 40×40 grid. If a user changes `width` or `height` via the sliders, the agents land outside valid coordinates or at unintended relative positions. Consider deriving positions proportionally from grid dimensions.

---

**`rng` in `model_params` uses `InputText`**

The `rng` parameter is exposed as a free-text input. The model does handle the string-to-int conversion with `contextlib.suppress(ValueError)`, but if a user types a non-integer string it silently falls back to `rng=None` (random seed). This should be a `SliderInt` or at least have a note in the label.

---

**README abbreviations unexplained**

Belief/desire keys like `Not-eh`, `bs`, `pa`, `wr`, `eh`, `w` come from the NetLogo model but are never decoded in the README. Adding a one-line legend (e.g., `pa = physical activity`, `bs = blood sugar`) would make the model interpretable without reading the original NetLogo source.

---

**Summary**

The core BDI logic is correct and the `mesa_signals` integration is a genuine strength. The two issues worth fixing before merge are: (1) replace `random.random()` with `self.model.random.random()` for reproducibility, and (2) fix the trust check in `evaluate_proposal` to use Bob's own trust rather than the sender's. The hardcoded coordinates are a minor usability issue. Great concept overall.

Harshini2411:
Cool example, BDI is a nice addition to the repo!

Abhinavk0220 already caught the big ones, just want to +1 the two that really need fixing before merge:

- **random.random() should be self.model.random.random()** since running with the same seed won't give reproducible results, which breaks a core expectation for Mesa examples. 
- **Trust check in evaluate_proposal uses the sender's trust instead of Bob's.** The doctor's trust always initialises above 0.5 so the condition is effectively always true and Bob's own self.trust never influences anything. 
- Also the **callable(a.goals) fix in the DataCollector** that ShreyasN707 flagged, looks like you said you'd update it but I don't see a new commit for that yet?

Otherwise the **mesa_signals / computed_property** usage is a genuine highlight here, good stuff!

champ-byte:
I have updated the trust . The  current implementation does give reproducible results , so no change should be needed, and  there is no code like `self.compliant = random.random() < self.model.compliance_rate` in the file

catherinedevlin:
I began some suggested refactoring at https://github.com/catherinedevlin/mesa-examples/tree/bdi_recommender, but on second thought, I don't think this is a good example of agent-based modeling.  The doctor, gym, etc. agents aren't actually active; they just represent possible decisions that "Bob" could make.  "Bob" is the only active *agent*; he really doesn't interact with anything but the user-set parameters; he makes a single decision and then moves to that.

Also, I wasn't able to vary his decision no matter what settings were used (he always moves to Doctor along the same path).

Perhaps there were plans to make the other agents into genuinely active agents in the future, but that doesn't seem to have happened yet.

champ-byte:
Hi, Yes Bob will always move towards the doctor. The original Netlogo version was designed that way and I had tried to almost recreate that logic.
The number of locations that bob has to visit will change if you keep PA=WR >=0.8. You can observe this by increasing the Proposal tick.

Yes even I think this might not be a good example for agent based modelling, but it is a good simple example to learn BDI

PR #391, opened 2026-03-16
fix(el_farol): update test to use num_agents instead of N

annapurna-gupta:
### Summary
While checking the el_farol example in mesa-examples, I noticed the test was failing because the ElFarolBar constructor expects `num_agents`, but the test was passing `N`. This PR updates the test to use the correct parameter.

### DIFF
--- a/examples/el_farol/tests.py
+++ b/examples/el_farol/tests.py
@@ -9,7 +9,9 @@ def test_convergence():
     # Testing that the attendance converges to crowd_threshold
     attendances = []
     for _ in range(10):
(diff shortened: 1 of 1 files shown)


ShreyasN707 (commented):
The changes of test parameters made in this PR is correct.
## Is it correct and up-to-date?

- [x] Uses current Mesa APIs  

- [x] The fix is minimal

## Is it clean and maintainable?

- [x] Minimal implementation  

- [x] Clear and readable 

- [x] Documentation is sufficient  
      (README explains what the model does, what it demonstrates, and how to run it)

EwoutH (approved):

EwoutH:
Clean and simple, thanks both!

PR #393, opened 2026-03-17
docs: add requirements.txt for termites example

tejas243:
The termites example was missing a requirements.txt 
file. I noticed this while running the examples 
locally  without it, users don't know what to 
install and the example fails to run.

Added mesa[viz]>=3.0 as the dependency since 
the example uses mesa visualization components.

Tested locally after adding the file.


quaquel:
You opened many PRs, none with a clear motivation nor agreement that these changes are needed.

PR #423, opened 2026-03-21
Add metadata forest fire

Vanya-kapoor:
Implements the metadata standard agreed upon in #327

Adds `metadata.toml`  to the forest_fire example with all required fields:
title, abstract, authors, domain, space, time, complexity, keywords, and
mesa version compatibility.

This serves as the reference implementation for the metadata standard
that will be applied to all examples.

Related: #3272
@jackiekazil @EwoutH @quaquel


codebreaker32:
1. clean your git history
2. Fix your description by explicitly mentioning the correct discussion
3. I don't think there has been yet a consensus

himanksingh2024-svg (commented):
Checked out the branch and ran solara run app.py locally — the forest fire visualization loads correctly, controls (Step, Play, Reset, Tree density slider) all work as expected.
On metadata.toml: Fields are comprehensive and well-structured. Note: Ewout has since proposed README frontmatter in #417 as the preferred direction (one file instead of separate metadata + README). Worth discussing whether to pivot to that format before this merges.
Unrelated change: The PR also modifies deffuant_weisbuch/model.py (replacing agents.to_list() with list(agents)) which is unrelated to forest_fire metadata. This should be a separate PR to keep changes focused and reviewable.
Overall the forest_fire example runs well and the metadata fields are solid. Main question is format alignment with #417.

Vanya-kapoor:
ok thanks a lot for your feedback!

Issue #357, opened 2026-03-05
Fix hex_ant: relative import error prevents example from running


Harshini2411:
## Describe the bug
The hex_ant example cannot be run due to a relative import error in app.py.

## To Reproduce
1. cd examples/hex_ant
2. python app.py

## Error
Traceback (most recent call last):
  File "app.py", line 6, in <module>
    from .agent import AntState
ImportError: attempted relative import with no known parent package

## Expected behavior
The example should run without errors.

## Environment
- Mesa version: 3.5.0
- Python version: 3.x
- OS: Windows

## Possible fix
Replace relative imports with absolute imports, similar to how other working examples in the repo are structured.

PR #427, opened 2026-03-21
fix: resolve relative import errors in emperor_dilemma, hex_ant, hex_…

Vanya-kapoor:
## Summary

Fixes `ImportError: attempted relative import with no known parent package`
in three examples that were broken on Mesa 3.3.1.

**Examples fixed:** emperor_dilemma, hex_ant, hex_snowflake
**Change:** `from .module import X` → `from module import X` (3 files, 1 line each)

## Root cause

These examples use relative imports (`from .agents import X`) which require
the directory to be a Python package with a parent context. Running
`solara run app.py` from inside the example directory doesn't provide that
package context, so the import fails. Changing to absolute imports fixes
this without requiring `__init__.py` files or changes to how the examples
are run.

## Verified

- `emperor_dilemma` — import fix confirmed, model initializes correctly ✓
- `hex_snowflake` — import fix confirmed, model initializes correctly ✓
- `hex_ant` — import fix confirmed, but exposes a secondary bug:
  `TypeError: 'PropertyLayer' object does not support item assignment`
  This is a separate Mesa 3.x API change unrelated to the import fix.
    
  
###  **GSoC contributor checklist**
**Context & motivation**
I was auditing all the examples of mesa 3.0 and found hex_ant,hex_snowflake have ImportError: relative import with no known parent package. That's what I tried to fix

**What I learned**
I learned that the dot in from .model import X tells Python to look for the module relative to a parent package — but when you run solara run app.py directly from inside the example folder, Python has no parent package context, so it crashes.  I also found that hex_ant had a second separate bug underneath the import error — a PropertyLayer API change 
Learning repo
🔗 **My learning repo**: <https://github.com/Vanya-kapoor/GSoC-learning-space/tree/main>
### Readiness checks

- [x] This PR addresses an agreed-upon problem (proposed and discussed in #417)
- [x] I have read the [contributing guide](https://github.com/mesa/mesa/blob/main/CONTRIBUTING.md) and [deprecation policy](https://github.com/mesa/mesa/blob/main/CONTRIBUTING.md#deprecation-policy)
- [x] I have performed a self-review: reviewed the PR diff and left inline comments on anything that needs explanation
- [x] Another GSoC contributor has reviewed this PR: @B2prakash 
- [x] No code changes — `pytest` and `ruff` checks not applicable
- [x] Documentation updated: this PR *is* the documentation addition
@jackiekazil @EwoutH

### DIFF
--- a/examples/emperor_dilemma/model.py
+++ b/examples/emperor_dilemma/model.py
@@ -1,10 +1,9 @@
 import random
 
+from agents import EmperorAgent
 from mesa import Model
 from mesa.datacollection import DataCollector
 from mesa.discrete_space.grid import OrthogonalMooreGrid
-
-from agents import EmperorAgent
 
 
 class EmperorModel(Model):
--- a/examples/emperor_dilemma/model.py
+++ b/examples/emperor_dilemma/model.py
@@ -4,7 +4,7 @@ from mesa import Model
 from mesa.datacollection import DataCollector
 from mesa.discrete_space.grid import OrthogonalMooreGrid
 
-from .agents import EmperorAgent
+from agents import EmperorAgent
 
 
 class EmperorModel(Model):
--- a/examples/hex_ant/model.py
+++ b/examples/hex_ant/model.py
@@ -1,7 +1,7 @@
 import mesa
 from mesa.discrete_space import HexGrid
 
-from .agent import Ant
+from agent import Ant
 
 
 class AntForaging(mesa.Model):
@@ -57,9 +57,9 @@ class AntForaging(mesa.Model):
         # Create the Nest in the center
         center = (self.grid.width // 2, self.grid.height // 2)
         # Spike the 'home' pheromone at the nest so ants can find it initially
-        self.grid.pheromone_home[center] = 1.0
+        self.grid.pheromone_home.data[center] = 1.0
         # Mark the home location
-        self.grid.home[center] = 1
+        self.grid.home.data[center] = 1
 
         # Scatter some Food Sources
         # Create 3 big clusters of food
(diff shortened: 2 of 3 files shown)


B2prakash (commented):
Good catch on the relative import issue. The fix is clean — one line per file, minimal change, clear root cause explanation.
A few comments:

1. The fix is correct — changing from .module import X to from module import X is the right approach for examples that are run standalone with solara run app.py. Since these aren't installed packages, relative imports won't work without parent package context.

2. Nice work documenting the hex_ant secondary bug. Flagging the PropertyLayer item assignment error separately rather than trying to fix it in this PR is good practice — keeps the PR focused on one issue.

3. GSoC checklist note: The "Another GSoC contributor has reviewed this PR" field is blank — I can fill that role. Consider adding my username there after this review.

4. One question: Did you check if any other examples in the repo have the same relative import pattern? If so, might be worth listing them in the PR description so someone can fix those too (in a separate PR).

Vanya-kapoor:
thanks for reviewing my pr. In reply to this question-Did you check if any other examples in the repo have the same relative import pattern? Yes I did, only these 4 had the same issue. 
This pr is ready to be merged. @jackiekazil @catherinedevlin @EwoutH

dietcoke-17 (changes requested):
Import fixes are correct and minimal — good. But hex_ant still crashes after this merge with TypeError: 'PropertyLayer' object does not support item assignment in model.py. fix the PropertyLayer API usage so hex_ant actually runs end-to-end. Closes #357 for real.

falloficaruss:
The `app.py` files need the same treatment for the examples to actually run with `solara run app.py`.

PR #443, opened 2026-03-26
Extend Emperor's Dilemma with WhistleblowerAgent and belief gap metric

TheMS1314:
## Summary

Extends the existing Emperor's Dilemma model with three additions:

**WhistleblowerAgent** — a new agent subclass immune to social pressure 
that always acts on private belief publicly, regardless of neighbour 
enforcement. Implemented as a clean subclass of EmperorAgent.

**Belief Gap metric** — tracks the average distance between private 
belief and public behaviour across all agents at each step. Measures 
the illusion of consensus directly.

**Norm collapse detection** — records the exact step when compliance 
first drops below 50% via model.collapse_step.

Setting fraction_whistleblowers=0.0 reproduces the original model exactly.

## Key finding

There is a sharp tipping point between 5-10% whistleblowers. Below this 
threshold, the norm holds completely. Above it, the norm collapses rapidly. 
This threshold behaviour emerges from local interaction rules, not from 
anything explicitly programmed.

## Changes

- agents.py — added WhistleblowerAgent subclass and belief_gap property
- model.py — added fraction_whistleblowers parameter, Belief Gap, and 
  Whistleblowers Defying metrics, collapse_step tracking
- app.py — added whistleblower slider, second chart for belief gap, 
  red color for whistleblower agents
- README.md — added extensions section documenting the new additions

---

### Context & motivation
I got drawn into the Emperor's Dilemma model because it connects to 
something I think about philosophically — how societies end up loudly 
defending things nobody actually believes in. The model formalizes that 
intuition. I wanted to extend it in a way that asks a concrete question: 
What actually breaks the trap? The whistleblower agent is my answer to 
that question.

### What I learned
The most surprising thing was the sharp tipping point behaviour — at 2% 
Whistleblowers, the norm holds completely, at 10% it collapses entirely. 
That threshold behaviour wasn't designed in; it emerged from the local 
interaction rules. I also learned why relative imports break with 
solara run directly and how to fix it. That took longer than expected.

### Learning repo
🔗 My learning repo: https://github.com/TheMS1314/GSOC_Learning_space
🔗 Relevant model: https://github.com/TheMS1314/GSOC_Learning_space/tree/main/models/emperor_dilemma

### Readiness checks
- [x] This PR addresses an agreed-upon problem (linked issue or discussion with maintainer approval), **or** is a small/trivial fix
- [x] I have read the contributing guide and deprecation policy
- [x] I have performed a self-review
- [ ] Another GSoC contributor has reviewed this PR: (pending)
- [x] Tests pass locally
- [x] Code is formatted (ruff check . --fix)
- [x] Documentation updated


examples/emperor_dilemma/agents.py
```diff
@@ -2,63 +2,96 @@
 
 
 class EmperorAgent(FixedAgent):
-    """An agent in the Emperor's Dilemma model.
+    """A citizen in the Emperor's Dilemma model.
```
TheMS1314, line 5: Renamed to "citizen" to make the social role explicit — these are people embedded in a society, not generic simulation agents.

examples/emperor_dilemma/agents.py
```diff
@@ -2,63 +2,96 @@
...
 
-    Inherits from FixedAgent because citizens do not move.
+    Inherits from FixedAgent because citizens do not move — they're embedded
+    in a social structure (grid or network) and only influence their neighbors.
+
+    Each agent has a private belief they keep to themselves and a public
+    behavior they show the world. These two things can be — and often are —
+    completely different. That gap is the whole point of the model.
```
TheMS1314, line 12: Added the private vs public behavior explanation directly in the docstring because that gap is the entire premise of the model. Anyone reading the code for the first time should understand the concept before reading the logic.

examples/emperor_dilemma/agents.py
```diff
@@ -2,63 +2,96 @@
...
-            model (mesa.Model): The model instance.
-            private_belief (int): The agent's private belief (-1 or 1).
-            conviction (float): The strength of the agent's conviction.
-            k (float): The cost of enforcement.
-        """
         super().__init__(model)
         self.private_belief = private_belief
         self.conviction = conviction
         self.k = k
+        self.agent_type = "citizen"
```
TheMS1314, line 20: Added agent_type so the visualization and DataCollector can distinguish citizens from whistleblowers without using isinstance checks.

examples/emperor_dilemma/agents.py
```diff
@@ -2,63 +2,96 @@
...
         self.compliance = self.private_belief
         self.enforcement = 0
 
     def step(self):
-        """Executes one step of the agent.
+        neighbors = self._get_neighbors()
+        num_neighbors = len(neighbors)
+
+        if num_neighbors == 0:
+            return
```
TheMS1314, line 30: Extracted neighbor fetching into a helper method so both EmperorAgent and WhistleblowerAgent reuse the same logic without duplication.

examples/emperor_dilemma/agents.py
```diff
@@ -2,63 +2,96 @@
...
+
+    def _get_neighbors(self):
         if self.cell is not None:
-            neighbors = list(self.cell.neighborhood.agents)
+            return list(self.cell.neighborhood.agents)
+        return []
+
+    @property
+    def belief_gap(self):
+        return 0 if self.compliance == self.private_belief else 1
```
TheMS1314, line 63: One line that captures the model's core concept — returns 1 if the agent is publicly lying about their belief, 0 if they're being honest. Used by compute_belief_gap in model.py.

examples/emperor_dilemma/agents.py
```diff
@@ -2,63 +2,96 @@
...
+
+
+class WhistleblowerAgent(EmperorAgent):
+    """A citizen who has decided they're done pretending.
+
+    Always acts on private belief publicly regardless of social pressure.
```
TheMS1314, line 69: Implemented as a subclass rather than a flag on EmperorAgent because the behavioral difference is fundamental — whistleblowers skip the pressure calculation entirely, which is a different kind of agent, not just a parameter tweak.

examples/emperor_dilemma/agents.py
```diff
@@ -2,63 +2,96 @@
...
-                self.enforcement = -self.private_belief
-            elif (self.compliance == self.private_belief) and (
-                (self.conviction * w_i) > self.k
-            ):
-                self.enforcement = self.private_belief
-            else:
-                self.enforcement = 0
+
+        # always act on private belief — pressure doesn't change this
+        self.compliance = self.private_belief
```
TheMS1314, line 84: This is the single line that defines a whistleblower. Compliance is set unconditionally before any pressure calculation runs. The enforcement logic below still executes normally.

examples/emperor_dilemma/app.py
```diff
@@ -6,14 +6,14 @@
     make_plot_component,
 )
 from mesa.visualization.components.portrayal_components import AgentPortrayalStyle
+from model import EmperorModel
```
TheMS1314, line 9: Changed from relative to absolute import because solara run executes app.py as a standalone script, not as part of a package — relative imports have no parent package to resolve against in that context.

examples/emperor_dilemma/app.py
```diff
@@ -6,14 +6,14 @@
...
-COLOR_COMPLY_QUIET = "#F0F8FF"  # AliceBlue
-COLOR_DEVIATE_QUIET = "lightgray"  # Light Gray
-COLOR_COMPLY_ENFORCE = "dimgray"  # Dark Gray
-COLOR_DEVIATE_ENFORCE = "black"  # Black
+# --- Colors ---
+COLOR_COMPLY_QUIET = "#F0F8FF"
+COLOR_DEVIATE_QUIET = "lightgray"
+COLOR_COMPLY_ENFORCE = "dimgray"
+COLOR_DEVIATE_ENFORCE = "black"
+COLOR_WHISTLEBLOWER = "#E63946"
```
TheMS1314, line 16: Bright red makes whistleblowers immediately visible on the grid so you can watch resistance spread or get overwhelmed as the simulation runs.

examples/emperor_dilemma/app.py
```diff
@@ -26,6 +26,10 @@ def emperor_portrayal(agent):
         zorder=1,
     )
 
+    if agent.agent_type == "whistleblower":
+        portrayal.update(("color", COLOR_WHISTLEBLOWER))
+        return portrayal
```
TheMS1314, line 31: Early return after setting whistleblower color so the compliance/enforcement color logic below doesn't run for them — whistleblowers always show red regardless of their enforcement state.

examples/emperor_dilemma/app.py
```diff
@@ -45,6 +49,12 @@ def post_process_lines(ax):
     ax.set_ylabel("Rate")
 
 
+def post_process_gap(ax):
+    ax.legend(loc="center left", bbox_to_anchor=(1, 0.9))
+    ax.set_ylabel("Belief Gap (0=honest, 1=mask)")
+    ax.set_ylim(0, 1)
+
```
TheMS1314, line 56: Separated into its own chart with y-axis fixed at 0-1 so it's easy to compare belief gap across runs at different whistleblower fractions without the scale shifting.

examples/emperor_dilemma/app.py
```diff
@@ -54,9 +64,16 @@ def post_process_lines(ax):
     post_process=post_process_lines,
 )
 
+belief_gap_component = make_plot_component(
+    {
+        "Belief Gap": "tab:purple",
+        "Whistleblowers Defying": "tab:orange",
+    },
+    post_process=post_process_gap,
+)
```
TheMS1314, line 73: Purple for Belief Gap, orange for Whistleblowers Defying — kept visually distinct from the green/red/blue of the compliance chart so the two charts are easy to read independently.

examples/emperor_dilemma/app.py
```diff
@@ -79,7 +96,8 @@ def post_process_space(ax):
         "values": [True, False],
         "label": "Cluster Believers (Homophily)?",
     },
-    "width": 40,
+    "fraction_whistleblowers": Slider("Fraction Whistleblowers", 0.0, 0.0, 0.5, 0.01),
+    "width": 25,
```
TheMS1314, line 100: Slider range 0.0 to 0.5 covers the interesting range. Width changed from 40 to 25 to match height and keep the grid square — the original was non-square.

examples/emperor_dilemma/app.py
```diff
@@ -95,7 +113,7 @@ def post_process_space(ax):
 page = SolaraViz(
     model,
     renderer,
-    components=[lineplot_component, CommandConsole],
+    components=[lineplot_component, belief_gap_component, CommandConsole],
```
TheMS1314, line 116: Added as a second chart component so both the compliance dynamics and the belief gap are visible simultaneously during a run.

examples/emperor_dilemma/model.py
```diff
@@ -1,16 +1,21 @@
 import random
 
+from agents import EmperorAgent, WhistleblowerAgent
```
TheMS1314, line 3: Same absolute import fix as app.py — needed for solara run to work correctly outside a package context.

examples/emperor_dilemma/model.py
```diff
@@ -1,16 +1,21 @@
...
+    """The Emperor's Dilemma Model — extended.
+
+    Simulates how an unpopular norm sustains itself through fear and false
+    enforcement, even when the vast majority privately reject it.
 
-    Simulates the spread of an unpopular norm using the Emperor's Dilemma logic.
+    Extensions over the base model:
+    - Whistleblower agents who publicly defy the norm regardless of pressure
+    - Belief gap metric tracking the illusion of consensus over time
+    - Norm collapse detection — the step at which the norm loses its grip
```
TheMS1314, line 18: Listed the three extensions explicitly in the class docstring so the scope of changes is clear without having to read through the whole file.

examples/emperor_dilemma/model.py
```diff
@@ -20,25 +25,18 @@ def __init__(
         fraction_true_believers=0.05,
         k=0.125,
         homophily=False,
+        fraction_whistleblowers=0.0,
```
TheMS1314, line 28: Defaults to 0.0 so the base model behavior is fully preserved — setting this to zero reproduces the original exactly.

examples/emperor_dilemma/model.py
```diff
@@ -20,25 +25,18 @@ def __init__(
...
-        """
         super().__init__(rng=rng)
 
         self.width = width
         self.height = height
         self.fraction_true_believers = fraction_true_believers
         self.k = k
         self.homophily = homophily
+        self.fraction_whistleblowers = fraction_whistleblowers
+        self.collapse_step = None
```
TheMS1314, line 39: Initialized to None so it's easy to check after a run whether collapse occurred — if it's still None the norm held for the entire simulation.

examples/emperor_dilemma/model.py
```diff
@@ -49,6 +47,8 @@ def __init__(
                 "Compliance": compute_compliance,
                 "Enforcement": compute_enforcement,
                 "False Enforcement": compute_false_enforcement,
+                "Belief Gap": compute_belief_gap,
+                "Whistleblowers Defying": compute_whistleblowers_defying,
```
TheMS1314, line 51: Belief Gap measures the illusion of consensus population-wide. Whistleblowers Defying stays near 1.0 by design — useful mainly to confirm whistleblowers are behaving correctly during testing.

examples/emperor_dilemma/model.py
```diff
@@ -83,53 +76,65 @@ def init_agents(self):
         else:
             believer_coords = set(random.sample(all_coords, num_believers))
 
+        disbeliever_coords = [c for c in all_coords if c not in believer_coords]
+        num_whistleblowers = int(len(disbeliever_coords) * self.fraction_whistleblowers)
+        whistleblower_coords = set(
+            random.sample(disbeliever_coords, num_whistleblowers)
+        )
```
TheMS1314, line 83: Sampled only from disbeliever coordinates — true believers can't be whistleblowers since they genuinely support the norm. Whistleblowing only makes sense for someone who privately rejects it.

examples/emperor_dilemma/model.py
```diff
@@ -83,53 +76,65 @@ def init_agents(self):
...
-
-        Shuffles agents and activates them (random order), then collects data.
-        """
         self.agents.shuffle_do("step")
         self.datacollector.collect(self)
 
+        if self.collapse_step is None:
+            compliance = compute_compliance(self)
+            if compliance < 0.5:
+                self.collapse_step = self.steps
```
TheMS1314, line 108: Checked every step but only records the first crossing — collapse_step is None guard ensures it doesn't get overwritten if compliance briefly dips and recovers.

examples/emperor_dilemma/model.py
```diff
@@ -83,53 +76,65 @@ def init_agents(self):
...
     disbelievers = [a for a in model.agents if a.private_belief == -1]
     if not disbelievers:
         return 0
     return sum(1 for a in disbelievers if a.enforcement == 1) / len(disbelievers)
+
+
+def compute_belief_gap(model):
+    if not model.agents:
+        return 0
+    return sum(a.belief_gap for a in model.agents) / len(model.agents)
```
TheMS1314, line 133: Delegates to the belief_gap property on each agent, which keeps the computation logic in one place and makes it easy to change the definition of "gap" without touching model.py.

examples/emperor_dilemma/model.py
```diff
@@ -83,53 +76,65 @@ def init_agents(self):
...
+
+
+def compute_whistleblowers_defying(model):
+    whistleblowers = [a for a in model.agents if a.agent_type == "whistleblower"]
+    if not whistleblowers:
+        return 0
+    return sum(1 for a in whistleblowers if a.compliance != 1) / len(whistleblowers)
```
TheMS1314, line 140: Filters by agent_type rather than class — avoids importing WhistleblowerAgent into the reporter functions and keeps them loosely coupled.

examples/emperor_dilemma/README.md
```diff
@@ -22,4 +22,30 @@ pip install mesa matplotlib solara
...
+### Norm Collapse Detection
+The model records exactly which step compliance first drops below 50%,
+turning a visual observation into a measurable data point accessible
+via `model.collapse_step`.
+
+### Key Finding
+There is a sharp tipping point around 5-10% whistleblowers. Below this
+threshold the norm holds completely. Above it the norm collapses rapidly.
+This threshold behavior emerges from the local interaction rules rather
+than being explicitly programmed.
```
TheMS1314, line 51: Added as a separate section at the bottom so the original model description stays intact and the extensions are clearly separated from the base documentation.

TheMS1314 (commented):
Self-review: added inline comments explaining each change and the reasoning behind it.

PR #445, opened 2026-03-27
Add Task system proof-of-concept: ant colony with interruptible tasks

samuelmohel:
## What this adds

A proof-of-concept implementation of the Task system proposed in 
projectmesa/mesa#2526 (EwoutH).

Adds a model-agnostic `Task` and `TaskQueue` class with:
- Explicit duration (step-based)
- Priority queue - highest priority task always executes first
- Interruptible and resumable task state
- Pluggable reward functions (linear, threshold, exponential)

Demonstrated via an ant colony where cave-in pheromone signals 
interrupt ongoing dig tasks in real time.

## Files
- `tasks.py` - reusable Task/TaskQueue classes, zero dependency on ant logic
- `model.py` - AntColonyModel + AntAgent
- `app.py` - SolaraViz visualization
- `metadata.toml` - example metadata

## Related
- Closes/references: projectmesa/mesa#2526 (Tasks discussion)
- Related: projectmesa/mesa#2538 (Behavioral Framework GSoC 2026)

## Preview
shot


DashamiJituri:
Hi @samuelmohel — I've been working on a related prototype (PR #432, needs-based Wolf-Sheep) and your TaskQueue implementation answers a question I had been sitting with.
In my model I emulated duration by tracking ticks_remaining as a manual counter on the agent — which works but is fragile when the agent gets removed mid-task. Your interruptible flag handles that cleanly.
On your design question about integer step-counts vs DiscreteEventScheduler float time — I'd lean toward accepting both via a union type (int | float), where integer duration maps directly to tick counts and float duration is resolved against model.time. That way Task stays compatible with both the standard scheduler and DEVS without forcing the user to choose upfront.
Happy to test this against my Wolf-Sheep model if it would help validate the approach.

samuelmohel:
Thanks @DashamiJituri that's a clean solution. The int | float union with integer mapping to tick counts and float resolving against model time keeps the API simple while covering both schedulers. One thing I'd add: Task would need to store the start_time at activation so float-duration tasks can compute remaining = (start_time + duration) - model.time correctly when resumed after interruption otherwise resumable tasks with float duration lose their progress reference. Would be great to test against your Wolf-Sheep model, especially to see how the interruptible flag behaves when an agent is removed mid-task. I'll open a follow-up branch with the union type implementation.

Issue #337, opened 2026-02-25
Fix emperor_dilemma: relative imports and outdated SpaceRenderer API


mehaksalil:
The emperor_dilemma example is currently broken and cannot be run due to two issues:

1. Relative imports in app.py, model.py, and agents.py are incompatible with 
   how Solara loads example modules.

2. SpaceRenderer.setup_agents() no longer exists in the current Mesa API.

Both issues prevent the example from running at all.

mehaksalil:
ERROR:    Traceback (most recent call last):
  File "C:\Users\mehak\AppData\Local\Programs\Python\Python311\Lib\site-packages\starlette\routing.py", line 694, in lifespan
    async with self.lifespan_context(app) as maybe_state:
  File "C:\Users\mehak\AppData\Local\Programs\Python\Python311\Lib\site-packages\starlette\routing.py", line 571, in __aenter__
    await self._router.startup()
  File "C:\Users\mehak\AppData\Local\Programs\Python\Python311\Lib\site-packages\starlette\routing.py", line 673, in startup
    handler()
  File "C:\Users\mehak\AppData\Local\Programs\Python\Python311\Lib\site-packages\solara\server\starlette.py", line 620, in on_startup
    appmod.ensure_apps_initialized()
  File "C:\Users\mehak\AppData\Local\Programs\Python\Python311\Lib\site-packages\solara\util.py", line 345, in wrapper
    return_value = f()
                   ^^^
  File "C:\Users\mehak\AppData\Local\Programs\Python\Python311\Lib\site-packages\solara\server\app.py", line 542, in ensure_apps_initialized
    app.init()
  File "C:\Users\mehak\AppData\Local\Programs\Python\Python311\Lib\site-packages\solara\server\app.py", line 106, in init
    app = self._execute()
          ^^^^^^^^^^^^^^^
  File "C:\Users\mehak\AppData\Local\Programs\Python\Python311\Lib\site-packages\solara\server\app.py", line 153, in _execute
    routes = [solara.autorouting._generate_route_path(self.path, first=True, initial_namespace=initial_namespace)]
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #463, opened 2026-04-14
fix(emperor-dilemma): replace relative imports and fix unseeded random calls

abhinavk0220:
## Summary

Fixes two bugs preventing the Emperor's Dilemma example from running, as reported in #337.

### Bug 1 — Relative imports break Solara module resolution
`app.py` used `from .model import EmperorModel` and `model.py` used `from .agents import EmperorAgent`. Solara loads example modules in a flat namespace and cannot resolve relative imports. Fixed by switching to absolute imports (`from emperor_dilemma.model import ...`), consistent with how other examples in this repo are structured (e.g. `bank_reserves`).

### Bug 2 — Module-level `random` bypasses the seeded RNG
`model.py` imported the stdlib `random` module and called `random.sample()` and `random.uniform()` during agent initialisation. This means the `rng` parameter passed to `EmperorModel()` — and exposed as a slider in the UI — had no effect on the initial believer placement or conviction values, making runs non-reproducible. Fixed by replacing both calls with `self.random.sample()` and `self.random.uniform()`.

## Test plan

- [ ] `solara run app.py` starts without `ImportError`
- [ ] Agents render on the grid with correct colors (comply/enforce states)
- [ ] Changing the Random Seed slider and clicking Reset produces a different initial configuration
- [ ] Line plot shows Compliance / Enforcement / False Enforcement over time

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/examples/emperor_dilemma/app.py
+++ b/examples/emperor_dilemma/app.py
@@ -1,3 +1,4 @@
+from emperor_dilemma.model import EmperorModel
 from mesa.visualization import (
     CommandConsole,
     Slider,
@@ -6,8 +7,6 @@ from mesa.visualization import (
     make_plot_component,
 )
 from mesa.visualization.components.portrayal_components import AgentPortrayalStyle
-
-from .model import EmperorModel
 
 # Colors matching Figure 2
 COLOR_COMPLY_QUIET = "#F0F8FF"  # AliceBlue
--- a/examples/emperor_dilemma/model.py
+++ b/examples/emperor_dilemma/model.py
@@ -1,10 +1,7 @@
-import random
-
+from emperor_dilemma.agents import EmperorAgent
 from mesa import Model
 from mesa.datacollection import DataCollector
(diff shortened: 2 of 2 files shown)


sadrasabouri:
Hey @abhinavk0220 
I verify that the issue you're trying to solve is still there (Python 3.12.13 and Mesa 4.0.0a0). I reviewed your PR, and here are some comments that can enhance it:

- Currently, in `examples/emperor_dilemma/README.md` we have `solara run app.py` as the run command. This implicitly assumes that the user is running the example when the working directory is `examples/emperor_dilemma`. Therefore, your fix will result in `ModuleNotFoundError: No module named 'emperor_dilemma'`. My suggestion is to move the `agents.py` and `model.py` into an insider directory `emperor_dilemma` (similar to `bank_reserves`).
- nit: make atomic-scoped PRs. The two bugs being solved by this PR do not live in the same scope and could be two separate valid PR.

The rest looks good to me.

PR #380, opened 2026-03-12
Fix shape_example: migrate from ModularServer to SolaraViz (Mesa 3.x)

Harshini2411:
Partially addresses #358

## Changes

**shape_example:**
- Deleted server.py (old ModularServer approach)
- Created app.py using SolaraViz + make_space_component
- Updated Walker to extend CellAgent, added step() method
- Updated run.py
- Note: arrowHead not supported in Mesa 3.x, replaced with marker="^"

**charts:**
- Deleted server.py (old ModularServer approach)
- Created app.py using SolaraViz, make_space_component, make_plot_component
- Uses AgentPortrayalStyle to fix FutureWarning
- Uses solara.Reactive to fix render loop
- Updated run.py

Fixes #358

PR #377, opened 2026-03-12
Fix hex_ant and hex_snowflake: migrate to Mesa 3.x visualization API

Harshini2411:
Fixes #357

Partially addresses #358 (hex_snowflake only)
- shape_example and charts will be fixed in a separate PR

### hex_ant (Fixes #357)
- Changed relative import `from .agent import Ant` to absolute 
  import `from agent import Ant` in model.py

### hex_snowflake (Partially addresses #358)
- Deleted server.py (old ModularServer approach)
- Created app.py using SolaraViz + make_space_component
- Updated run.py to import from app.py
- Deleted portrayal.py (replaced by agent_portrayal in app.py)

## Testing
- Mesa version: 3.5.0
- Python: 3.x
- OS: Windows

PR #470, opened 2026-06-16
fix(hex_ant): fix relative imports and PropertyLayer API to make example fully runnable

### LINKED ISSUE #357: Fix hex_ant: relative import error prevents example from running

dietcoke-17:
### Summary
Closes #357 — hex_ant was completely non-functional. Running `solara run app.py` 
crashed immediately and the model never reached the browser. This PR works through 
every crash in sequence until the example fully runs.

### Bug / Issue
**Bug 1** — Relative imports in app.py and model.py

app.py — before
from .agent import AntState
from .model import AntForaging
 removed the . before the module

why this change
Relative imports (`.module`) only work when Python treats the directory as a package with a known parent. When Solara executes `app.py` directly, `__package__` is `None`, so there is no parent to be relative to

**Result:** Model loads past the import stage.

**Bug 2** — Direct PropertyLayer item assignment in `_init_environment`
model.py — before
self.grid.pheromone_home[center] = 1.0
self.grid.home[center] = 1

model.py — after
self.grid.pheromone_home.data[center] = 1.0
self.grid.home.data[center] = 1

**Why**
In Mesa 3.x, `PropertyLayer` is a wrapper object, not a numpy array. Direct item assignment (`layer[x, y] = val`) is not supported. The underlying numpy array is accessed through the `.data` attribute, which does support index assignment.

**Result:** Model initialises, nest and food sources are placed, 
visualization opens in browser.

**Bug 3** — Wrong dynamic layer lookup in `_update_pheromone_layer`
model.py — before
np_layer = self.grid.property_layers[layer_name]

model.py — after
np_layer = getattr(self.grid, layer_name).data

**Why:** 

self.grid.property_layers does not exist as an attribute. `HexGrid` overrides `__getattr__` and routes ALL attribute access through the internal property layer dictionary — so accessing `.property_layers` is itself treated as a layer name lookup and fails. The correct pattern for dynamic layer access by name is `getattr(self.grid, layer_name)`. The `.data` is also needed here for the same reason as Bug 2 — numpy operations like `*=` and boolean masking must be performed on the raw 
array, not the wrapper object.

### Testing

{F6415441-5EB7-41B7-B4DC-BCB88C82849E}

## Relation to existing PRs

- #377 and #380 — attempted import fixes only, both closed
- #427 — correctly fixes relative imports; the copy of the code worked on here already had clean absolute imports, so that was not the blocker. The PropertyLayer bugs are entirely separate and were not addressed by any existing PR.
- Bug 3 specifically would have crashed the model on step 1 even if every other fix had been applied first — it was the last hidden blocker.

## Verified
Tested on Windows, Mesa 3.5.0, Python 3.14:

- `solara run app.py` starts without error 
- Visualization opens at http://localhost:8765 
- Ants spawn at nest and begin foraging 
- Pheromone trails build up and evaporate each step 
- Returning ants (gold) follow home pheromone correctly 
- Food clusters deplete as ants pick up food 

**After Codex Suggestion**
In `model.py` and `app.py`: used try/except import pattern to support both `solara run app.py` (direct execution) and package import via test runner and it worked 

{335454A4-326F-42A9-8570-CFAF0F73B7DD}

**Also CI Failing Issue**
Also fixed a pre-existing lint error in rl/wolf_sheep/agents.py — two variables x, y were unpacked but never actually used anywhere. Renamed them to _x, _y to tell the linter they are intentionally unused. This was flagged by pre-commit CI and is unrelated to the hex_ant fix.

### DIFF
--- a/examples/hex_ant/app.py
+++ b/examples/hex_ant/app.py
@@ -3,8 +3,8 @@ from matplotlib.colors import LinearSegmentedColormap, ListedColormap
 from mesa.visualization import SolaraViz, make_space_component
 from mesa.visualization.components import PropertyLayerStyle
 
-from .agent import AntState
-from .model import AntForaging
+from agent import AntState
+from model import AntForaging
 
 plt.rcParams["figure.figsize"] = (10, 10)
 
--- a/examples/hex_ant/app.py
+++ b/examples/hex_ant/app.py
@@ -3,8 +3,12 @@ from matplotlib.colors import LinearSegmentedColormap, ListedColormap
 from mesa.visualization import SolaraViz, make_space_component
 from mesa.visualization.components import PropertyLayerStyle
 
-from agent import AntState
-from model import AntForaging
+try:
+    from .agent import AntState
+    from .model import AntForaging
+except ImportError:
+    from agent import AntState
+    from model import AntForaging
 
 plt.rcParams["figure.figsize"] = (10, 10)
 
--- a/examples/hex_ant/model.py
+++ b/examples/hex_ant/model.py
@@ -1,7 +1,10 @@
 import mesa
 from mesa.discrete_space import HexGrid
 
-from agent import Ant
+try:
+    from .agent import Ant
+except ImportError:
+    from agent import Ant
 
 
 class AntForaging(mesa.Model):
--- a/examples/hex_ant/model.py
+++ b/examples/hex_ant/model.py
@@ -1,7 +1,7 @@
 import mesa
 from mesa.discrete_space import HexGrid
 
-from .agent import Ant
+from agent import Ant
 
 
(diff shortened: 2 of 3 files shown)


tpike3 (approved):
Thanks @dietcoke-17

PR #474, opened 2026-07-26
fixed: resolve emperor_dilemma import errors completely (builds on #427)

dietcoke-17:
solara run app.py currently fails on this example with:

#427 addresses part of this by converting model.py's `from .agents import EmperorAgent` to an absolute import, but two problems remains:

**1. app.py has the identical relative import**
**2. test_examples.py imports every model.py via a dotted path** (examples.emperor_dilemma.model), which does have proper package context, so it needs the relative import to keep working.

**Fix**

Wrap both imports in try/except: try the relative import first (works when imported as part of the package, e.g. under pytest), fall back to the absolute import (works when Solara execs the file directly). Same pattern already used in `hex_ant/app.py`.

**Testing**
-`pytest test_examples.py -k emperor -v` → `test_model_steps[EmperorModel] PASSED`
-`ruff check examples/emperor_dilemma/` → `All checks passed!`
- `solara run app.py` from within `examples/emperor_dilemma/` 
→ starts cleanly, page returns HTTP 200, grid and compliance/enforcement/false-enforcement charts render correctly


dietcoke-17:
CI's build-pre job (Mesa installed from its GitHub main branch, 4.0.0a0) fails on AntForaging because PropertyLayer.data returns a memoryview there instead of a numpy.ndarray. Wrapping with np.asarray() gives a zero-copy view that works either way.

Verified against both mesa==3.5.1 and mesa 4.0.0a0.

tpike3 (changes requested):
@dietcoke-17 Thanks for doing these!!! We really appreciate the work. 

Could you remove the hex_ant fix for two reasons: 
1- This was fixed in #476 (its the same fix so it would be genrally fine to merge)
2 - We want to keep PRs (even if simple) confined to one problem in this case the emperor's dilemma

Thank you again for all your work

dietcoke-17:
Surely i thought it was problem in my pr so i updated it as per the requirement and after that pr got merged there is no update of this older pr as per new fix how could we do that like mesa repo here in mesa example it is a curious question for me?

dietcoke-17:
Also after this one we can close #427 i have fixed all three issue showed their separately in 3 prs

tpike3 (approved):
LGTM Thanks!

PR #477, opened 2026-07-27
fix: port hex_snowflake to SolaraViz (supersedes hex_snowflake part of #427)

dietcoke-17:
**Problem**

`hex_snowflake` can't run under any single Mesa version right now:

- `server.py` calls `mesa.visualization.ModularServer` / `CanvasHexGrid` — both removed in Mesa 3.x
- `model.py` / `cell.py` already use `mesa.discrete_space` (`HexGrid`, `FixedAgent`) — Mesa-3.x-only
- `requirements.txt` pins `mesa~=2.0`

With Mesa 3.x installed, `python run.py` fails:

**Relation to #427**

#427 proposes converting `model.py`'s `from .cell import Cell` to an absolute import, mirroring its emperor_dilemma/hex_ant fixes. I tested this and it doesn't apply here — `app.py` (below) imports the inner package via `from hex_snowflake.model import HexSnowflake`, the same pattern the existing `run.py` already used successfully. That gives `model.py` proper package context, so its relative import already resolves correctly as-is. Making it absolute would break it: `ModuleNotFoundError: No module named 'cell'`.

So this PR leaves `model.py`/`cell.py` completely untouched and fixes the actual blocker instead: the removed visualization API.

**Fix**

Replaced `run.py` + `server.py` + `portrayal.py` with a single `app.py` using `SolaraViz`, matching the structure already used by `hex_ant`/`emperor_dilemma`
Also removed `requirements.txt` (stale `mesa=2.0` pin — neither `hex_ant` nor `emperor_dilemma` pins a version) and updated `Readme.md`'s run instructions (`mesa runserver` → `solara run app.py`).

**Testing**

- `solara run app.py` from `examples/hex_snowflake/` → HTTP 200, no traceback
- `pytest test_examples.py -k Hex` → `test_model_steps[HexSnowflake] PASSED`
- `ruff check examples/hex_snowflake/` → `All checks passed!`

**Results**

Screenshot 2026-07-27 at 7 32 34 AM


tpike3:
Thanks @dietcoke-17 , just a friendly reminder please update your fork so it merges #476 and the tests will pass.

PR #436, opened 2026-03-24
Add rumor spread example with Solara visualization

RodionOm:
-- Summary

This PR adds a small Solara-based rumor spread example to 'mesa-examples'
d7e559d8-010d-4ff8-91a1-39cc187fdc83
80a6e29c-ef5b-4dd2-9b70-ea01adbdc7d9
2c46e799-ba70-48f4-8f12-79fe0c984b47
.

The model places agents on a 2D toroidal grid. A configurable number of agents initially know a rumor. Informed agents may spread the rumor to neighboring agents, while a recovery mechanism allows agents to forget the rumor and return to the uninformed state.

-- Features

- Solara-based interactive visualization
- Spatial grid view of rumor diffusion
- Time-series plot of informed agents
- Parameter controls for:
  - initial spreaders
  - infection strength
  - recovery rate

-- Behavior

The model demonstrates multiple regimes depending on parameter values:

- **Extinction**: low infection strength and/or high recovery lead the rumor to die out
- **Equilibrium**: balanced infection and recovery produce sustained fluctuations
- **Saturation**: high infection strength and low recovery lead to near-total spread

-- Motivation

I built this example while preparing for GSoC and learning Mesa hands-on. I wanted to create a compact, reproducible model that demonstrates parameter-dependent rumor diffusion behavior in a way that is easy to run and inspect.

-- Notes

This model is intentionally simple and designed as an educational example rather than a realistic social diffusion simulator.

PR #478, opened 2026-07-28
Added recovery_rate parameter to Rumor Mill model

### LINKED ISSUE #472: Add recovery rate in Rumor_mill model
This idea of adding recovery rate is initially got from the PR #436 which is based on the same concept as we have already a rumor_mill and we also can have network topology on it

dietcoke-17:
**Summary**

Added a `recovery_rate` parameter to the Rumor Mill example, letting informed agents forget the rumor and become uninformed again. Closes #472 

The model was purely monotonic -- `Percentage_Knowing_Rumor` could only trend upward, so the only observable outcome was "how fast does it saturate." Recovery/forgetting is a standard ingredient in rumor- and epidemic-spread models (the "R" side of SIR/SIS dynamics), and adding it unlocks three qualitatively different regimes:

- **Extinction** -- recovery outpaces spread, the rumor dies out
- **Equilibrium** -- spread and recovery balance, producing a fluctuating steady state
- **Saturation** -- spread outpaces recovery, near-total adoption

Related to #436 (a separate rumor-spread example that already modeled recovery) and #451 (which proposes a forgetting mechanic plus additional agent types for this same model) -- flagging both for reviewers since there's conceptual overlap. This PR is intentionally scoped to just the recovery-rate piece to keep it small and easy to review.

**Changes**

- `rumor_mill/agent.py`: `Person` gains `recovery_rate` and a `just_recovered` flag; `step()` checks for forgetting (probability `recovery_rate`) before attempting to tell a neighbor
- `rumor_mill/model.py`: `RumorMillModel` accepts `recovery_rate` (default `0.0`, so existing callers see no behavior change), threads it to every agent, and adds a `Percentage_Recovered_This_Step` reporter
- `app.py`: new `recovery_rate` slider (interactive demo defaults it to `0.05`)
- `README.md`: documents the parameter and the extinction/equilibrium/saturation regimes
- `tests.py`: new tests for backward compatibility, the deterministic full-recovery edge case, parameter wiring, and the new reporter

**Design Note**

Forgetting is implemented as SIS-style: an agent that forgets becomes fully susceptible again and can be told and re-adopt the rumor later, rather than SIR-style permanent immunity. That matches how #436 described its own recovery mechanic. Happy to switch if maintainers prefer immunity instead.


examples/rumor_mill/rumor_mill/agent.py
```diff
@@ -23,27 +23,32 @@ def __init__(self, model, cell, rumor_spread_chance=0.5, color=None):
...
+        if neighbors:
+           neighbor = self.random.choice(neighbors)
+           if (
+               not neighbor.knows_rumor
+               and self.random.random() < self.rumor_spread_chance
+           ):
+                neighbor.knows_rumor = True
+                neighbor.newly_learned = True
+        neighbor.times_heard += 1
+        neighbor.times_heard_this_step += 1
```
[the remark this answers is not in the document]
dietcoke-17: Good catch on the missing guard, thanks for flagging — re-indented times_heard/times_heard_this_step to stay inside the if neighbors: block, and added test_no_crash_with_empty_neighborhood (1x1 grid) to lock it in.

examples/rumor_mill/rumor_mill/agent.py
```diff
@@ -23,27 +23,32 @@ def __init__(self, model, cell, rumor_spread_chance=0.5, color=None):
...
+           agent for agent in self.cell.neighborhood.agents if agent != self
+        ]
+        if neighbors:
+           neighbor = self.random.choice(neighbors)
+           if (
+               not neighbor.knows_rumor
+               and self.random.random() < self.rumor_spread_chance
+           ):
+                neighbor.knows_rumor = True
+                neighbor.newly_learned = True
```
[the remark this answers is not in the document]
dietcoke-17: Good catch — newly_learned was only ever meaningful pre-recovery, when learning could happen at most once per agent. Fixed by adding a permanent has_known_rumor_before flag on Person; newly_learned (and so New_People_Knowing_Rumor) now only fires on an agent's genuine first adoption, never on re-adoption after forgetting. Added a regression test (test_recovered_agents_are_not_double_counted_as_new) that forces heavy recover/re-learn cycling and asserts the cumulative new-learner count stays under the population that started uninformed.

dietcoke-17:
Two issues surfaced during review of the recovery_rate addition:
- New_People_Knowing_Rumor was inflated whenever an agent forgot the rumor and was later re-told it, since newly_learned fired on every adoption rather than just the first. Fixed by adding a permanent has_known_rumor_before flag and only setting newly_learned the first time an agent ever adopts the rumor.
- Stepping an informed agent with an empty neighborhood (e.g. a 1x1 grid) raised UnboundLocalError, because the times_heard counter updates sat outside the `if neighbors:` block instead of nested inside it.
(FROM SELF REVIEW)
-Also added test recovered agents are not double counted as new and test no crash with empty neighborhood to cover both.

examples/rumor_mill/rumor_mill/model.py
```diff
@@ -16,6 +16,7 @@ def __init__(
         height=10,
         know_rumor_ratio=0.01,
         rumor_spread_chance=0.5,
+        recovery_rate=0.0,
         eight_neightborhood=False,
         rng=None,
```
falloficaruss: Existing positional calls such as `RumorMillModel(10, 10, 0.1, 0.5, True, 1)` will now interpret True as a recovery rate and 1 as the neighborhood flag, silently changing model behavior and losing the seed. You may want to take a look at the position of recovery_rate

falloficaruss (commented):
Thank you for the PR! 

You must not name added test files directly as `test.py`. There is also a breaking change which you should take a look at.

examples/rumor_mill/rumor_mill/model.py
dietcoke-17: Yep surely I didn't think about that lemme think about that again

dietcoke-17: just rechecked and moved the recovery_rate to the end of the parameter list (after rng) instead of inserting it in the middle, so existing positional calls map to the same arguments as before. Applied the same fix to Person.__init__ in agent.py. recovery_rate came back as 0.0 instead of True, and the seed still worked also added a test that reproduces our exact example call and asserts recovery_rate stays at its default and the seed still works.

falloficaruss: You should also change the name of test file to something like test_rumor_mill.py

tpike3 (approved):
These are some great improvements, thanks @dietcoke-17

PR #485, opened 2026-08-23
Add energy expenditure example

### LINKED ISSUE #446: Add energy-based agent example (needs-based behavior)
I would like to contribute a new example demonstrating
needs-based agent behavior using an energy model.

The example will include:
- Agents losing energy over time
- Decision-making based on energy levels (move vs rest)
- Simple, clear structure for educational purposes

This aligns with the behavioral modeling direction mentioned
in the GSoC ideas (needs-based behavior exploration).

Please let me know if this approach is acceptable.

lahari936:
Add a minimal needs-based behavior example: agents pay energy every step, pay more when they move, switch to resting once their reserve drops below a threshold, and are removed when it reaches zero.

Written against the current Mesa API (mesa.discrete_space, CellAgent, rng seeding) and laid out like the other examples in this repo, with a DataCollector, SolaraViz sliders and plots, and tests covering the energy and population invariants.

Starting energy is drawn per agent rather than shared, so the population declines over many steps instead of expiring all at once.

Closes #446

Thanks for opening a PR! Please click the `Preview` tab and select a PR template:

- [🐛 Bug fix](?expand=1&template=bug.md)
- [🛠 Feature/enhancement](?expand=1&template=feature.md)


falloficaruss:
Thank you for the PR!
Can you please describe more on the use of this model?

lahari936:
Thanks for taking a look!

The model is meant as the smallest complete demonstration of *needs-based
behavior* — an agent whose actions are selected by its own internal state
rather than by chance. That was the motivation in #446
<https://github.com/lahari936/mesa-examples/issues/446>, following the
behavioral-modeling direction in the GSoC ideas list.

Concretely, each agent carries one number, energy. Every step costs
metabolism just to exist; moving costs move_cost on top of that. Once
energy falls below rest_threshold the agent stops moving and rests. That
single comparison is the entire decision rule — there is no "try to
survive" instruction anywhere in the code — and yet what emerges is an
agent that trades mobility for lifespan exactly when it can no longer
afford to wander. That trade-off, appearing without being programmed, is
the thing the example is trying to show.

At the population level the three collected series make it visible: Alive,
Resting, and Mean Energy. Starting energy is drawn per agent rather than
shared, so instead of every agent expiring on the same step you get a
decline curve, with a clear wave where the population flips from
mostly-moving to mostly-resting before it dies off.

Where I think it earns a place next to the existing examples: Wolf-Sheep
also has energy, but there it is entangled with predation, reproduction,
and grass regrowth, so it is hard to point at energy alone and say "this is
what drove that behavior". Here energy is the only state variable, which
makes it usable as a teaching model and as a substrate — the threshold rule
sits in one method, so it can be swapped for a utility function, a needs
hierarchy, or a behavior tree while the grid, data collection, and
visualization stay untouched.

The deliberate limitation is that nothing replenishes energy, so extinction
is certain. That is intentional: it keeps the example to one concept. If
you would rather it demonstrate a steady state, adding a food source and a
foraging drive is a small change and I am happy to make it — likewise happy
to expand the example's README with the above if that framing is useful to
have in the repo rather than only in this thread.
------------------------------

lahari936 (commented):
Thanks! The model is the smallest complete demo of needs-based behavior: each agent carries one number, `energy`. Existing costs `metabolism` every step, moving costs `move_cost` on top, and below `rest_threshold` the agent stops moving and rests. That single comparison is the whole decision rule — nothing tells the agent to survive, yet it trades mobility for lifespan exactly when it can no longer afford to wander.

`Alive`, `Resting` and `Mean Energy` make that visible at the population level; starting energy is drawn per agent, so you get a decline curve with a clear wave from mostly-moving to mostly-resting rather than everyone expiring on the same step.

Wolf-Sheep has energy too, but tangled with predation, reproduction and grass regrowth. Here energy is the only state, so the mechanism stays legible and the threshold rule sits in one method — easy to swap for a utility function or behavior tree while the grid, data collection and viz stay untouched.

Nothing replenishes energy, so extinction is certain; that's deliberate, to keep it to one concept. Happy to add a food source and foraging drive if you'd rather it show a steady state, or to fold this explanation into the example's README.

falloficaruss:
Using LLMs for writing code is fine but you must not just act as a proxy between the reviewer and the LLM.
You may want to read: [Guideline on AI and LLM usage](https://github.com/mesa/mesa/blob/main/CODE_OF_CONDUCT.md#ai-and-llm-use)

Nevertheless going through the example I think needs based behavior is already covered by Humanitarian Aid Distribution model.

lahari936:
Thanks for the feedback!

